PR #41, opened 2023-12-20
feat: compress proof image before upload

raphodn:
### What

Use [compressor.js](https://github.com/fengyuanchen/compressorjs) to compress the proof image before uploading it to the backend.

Tested with 7.1 MB JPEG image --> 677 KB output

### DIFF
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -12,7 +12,6 @@
           <v-divider></v-divider>
           <v-card-text>
             <v-file-input
-              show-size
               accept="image/*"
               label="Proof"
               v-model="proofImage"
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -136,9 +136,18 @@
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


src/views/AddPriceSingle.vue
```diff
@@ -136,10 +135,19 @@
...
+import Compressor from 'compressorjs'
 import api from '../services/api'
 import BarcodeScanner from '../components/BarcodeScanner.vue'
 import LocationSelector from '../components/LocationSelector.vue'
 
+Compressor.setDefaults({
+  checkOrientation: true,  // default
+  retainExif: true,
+  quality: 0.6,
+  // mimeType: 'image/webp',
```
raphodn, line 147: @raphael0202 I commented it out because I have issues when the config is set to webp: the image is compressed, and the output is displayed, but it stored without an extension (None instead). It's probably due to the backend image mgmt, so we'll need a seperate PR to fix it ;)

raphodn: done here ! https://github.com/openfoodfacts/open-prices-frontend/pull/44

PR #49, opened 2023-12-22
feat: price create form: improve a bit the proof upload (responsive, overflow, icon)

raphodn:
### What

Proof upload form changes:
- change icon (square -> circle)
- avoid overflow
- responsive : display image on the right column on bigger screens
- improve loading (hide error, show success)

### DIFF
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -7,20 +7,30 @@
         <v-card
           title="Take a picture of the price tag"
           subtitle="We need this for proof :)"
-          prepend-icon="mdi-numeric-1-box-outline"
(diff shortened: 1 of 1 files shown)


teolemon:
During my days off, I took the time to try the UI. Something that would be awesome on proof upload would be to attempt to extract a barcode, and prefill the barcode field.

raphodn:
> During my days off, I took the time to try the UI. Something that would be awesome on proof upload would be to attempt to extract a barcode, and prefill the barcode field.

Yes definitely ! We talked about it on wednesday, can you create an issue in the backend repo ? the server could answer with info such as the barcode, the price, the product_name...

PR #67, opened 2023-12-25
fix: replace cookies with localstorage

raphodn:
### What

We currently use cookies to store username & token. But :
- they seem to expire quite quickly (need to re-login often)
- we also use the localstorage to store other info

TODO : make localstorage interactions reactive ?


raphodn:
see #69 instead

PR #69, opened 2023-12-26
refactor: replace cookie storage with a store (pinia)

raphodn:
### What

We currently use cookies to store username & token. But they seem to expire quite quickly (need to re-login often).

It's a good opportunity to replace this with a store : [pinia](https://pinia.vuejs.org/)

Next step: persist the store in the localstorage

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "compressorjs": "^1.2.1",
     "html5-qrcode": "^2.3.8",
     "leaflet": "^1.9.4",
+    "pinia": "^2.1.7",
     "universal-cookie": "^6.1.1",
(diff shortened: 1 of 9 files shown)


raphodn:
Is localStorage less secure than cookie ? for instance this article : https://dev.to/rdegges/please-stop-using-local-storage-1i04
- see the comments : cookies are very limited in size, hard to manage
- localStorage is vulnerable to XSS attack, depends on scripts that are imported
- there's always the possibility to revert token storage to cookies, BUT need also to implement cookie authentication + httpOnly (was using bearer token for now)
- idea of setting an expiry date to tokens (but the user will need to sign in again ??)

PR #26, opened 2023-12-18
feat: simple list of last added prices

raphodn:
### What

- New nav drawer item "Last prices"
- show most recent prices first
- simple card with price / product / location details (dedicated component)

### Todo (in futur PRs)

- currency : avoid hardcoded "€"
- quantity : display products with "L"
- bigger image : 2 columns instead of "card"
- search, reorder, filter

### Questions

- what to display if information missing ?

### Screenshot

image

PR #77, opened 2023-12-27
feat(price card): improve look & feel

raphodn:
### What

Price card v2 (v1 in #26)

Changes : 
- columns
- chips for basic info
- click on username to go to user page (thanks to #78 )

### Screenshot

image

### DIFF
--- a/src/components/PriceCard.vue
+++ b/src/components/PriceCard.vue
@@ -2,14 +2,14 @@
   <v-card>
     <v-container class="pa-2">
       <v-row>
-        <v-col style="max-width:25%">
+        <v-col v-if="!hideProductImage" style="max-width:25%">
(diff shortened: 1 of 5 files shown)


raphael0202 (approved):
Nice revamp :+1:

teolemon:
I got some feedback that the store indication could be ambiguous. We already have two Auchan in Grenoble. We'll need to see if we want to act on this (perhaps with hash-based colors ?).

raphodn:
> I got some feedback that the store indication could be ambiguous. We already have two Auchan in Grenoble. We'll need to see if we want to act on this (perhaps with hash-based colors ?).

Yes i was thinking of : 
- adding a :fr: flag instead of the 📍 to show the location
- adding the road info, but : it could make the text quite long ; we don't store it yet in the backend..

PR #117, opened 2024-01-07
feat(location detail): change url (osm type & id instead of id)

raphodn:
### What

Thanks to https://github.com/openfoodfacts/open-prices/pull/108 we can now retrieve product info by code.

This change will simplify and clarify urls : 
- before `/locations/1`
- after `/locations/node/652825274`

Also improve display of unknown location (show card with location type & id)

### DIFF
--- a/src/components/PriceCard.vue
+++ b/src/components/PriceCard.vue
@@ -213,7 +213,7 @@ export default {
       if (this.readonly) {
         return
       }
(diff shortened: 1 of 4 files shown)


raphodn:
Kinda linked to #1553

The problem with this PR, is that it doesn't take into account non-OSM locations (online).

PR #141, opened 2024-01-13
feat(stats): store results instead of fetching data each time

raphodn:
### What

WIP


Valimp:
@raphodn Which part do you need help with?

PR #160, opened 2024-01-17
feat(i18n): add languages de nl it pt and update es fr

dq18:
### What
- Update a few languages


raphodn (approved):
top ! :globe_with_meridians:

PR #174, opened 2024-01-18
fix(crowdin): change config to simplify PR from crowdin

dq18:
### What
- Replace commit message from crowdin with the same message. This should simplify a squash to only display this message
- Added [ci skip] tag to not trigger a build for each commit

https://developer.crowdin.com/configuration-file/


raphodn (approved):
:+1:

PR #241, opened 2024-01-26
feat(i18n): Add new languages

dq18:
### What
- added all languages from Crowdin to the app
- added a translation completion level and number display in settings

List of languages was modified from:
https://github.com/annexare/Countries/blob/main/packages/countries/src/data/languages.ts


src/i18n/data/languages.json
```diff
@@ -0,0 +1,937 @@
+[
+  {
+    "code": "aa",
+    "name": "Afar",
+    "native": "Afar"
+  },
+  {
+    "code": "ab",
+    "name": "Abkhazian",
+    "native": "\u0410\u04a7\u0441\u0443\u0430"
```
raphodn, line 10: why the special characters ? not possible to avoid them ?

src/i18n/localeManager.js
```diff
@@ -78,8 +79,66 @@ const localeManager = {
...
+        const localeCode = key.replace('./locales/', '').replace('.json', '')
+        return localeCode;
+      })
+  
+      return locales;
+    } catch (error) {
+      console.error('Error fetching locales:', error);
+    }
+  },
+  async calculateTranslationCompletion(locale) {
```
raphodn, line 97: complex method, needs a description / docstring

src/views/UserSettings.vue
```diff
@@ -69,23 +73,41 @@ export default {
...
 
+  },
   computed: {
     ...mapStores(useAppStore),
     formFilled() {
       return Object.values(this.userSettingsForm).every(x => !!x)
     },
   },
-  mounted() {
+  async mounted() {
```
raphodn, line 97: "async" isn't used anywhere, are you sure we need it here ?

src/views/UserSettings.vue
```diff
@@ -69,23 +73,41 @@ export default {
...
+      languageList: [],
+      languagesData: null,
+      languageTranslationCompletion: null,
     }
   },
+  watch:{
+    'userSettingsForm.selectedLanguage': async function () {
+      console.log("selected language: ", this.userSettingsForm.selectedLanguage)
+      if (this.userSettingsForm.selectedLanguage !== null) {
+        this.languageTranslationCompletion = await localeManager.calculateTranslationCompletion(this.userSettingsForm.selectedLanguage.code)
```
raphodn, line 85: in the v2 it could be a github action instead (that stores the completion value in the `languages.json`)

raphodn (approved):
looks good ! don't forget to cleanup the console.logs :)

src/views/UserSettings.vue
dq18: Yes definitely

src/views/UserSettings.vue
dq18: in line 99 I put
`    const { default: languagesData } = await import('@/i18n/data/languages.json')
`

src/i18n/data/languages.json
dq18: I updated the file (formatting issue on Windows)

src/i18n/localeManager.js
dq18: Done

PR #246, opened 2024-01-27
docs: improve setup instructions

alexfauquette:
I propose to move the instruction directly in the README.md to avoid contributor missing it. For now, we don't miss places on this document so no need to split

Added a section for newbies to give them correct tools, plus a direct link to the server project

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,15 +2,19 @@
 
 ## Installation
 
-```sh
-yarn
-```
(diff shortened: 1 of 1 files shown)


raphodn:
> I propose to move the instruction directly in the README.md to avoid contributor missing it

there was already a link to `CONTRIBUTING.md` :thinking: 
I'd rather keep it that way, similar to https://github.com/openfoodfacts/open-prices and many other projects :)

raphodn (approved):
thanks for your first contribution :pray:

Issue #65, opened 2023-12-25
Papercut tracker


teolemon:
### What

```[tasklist]
### Papercuts
- [x] The take a picture button should be the Camera plus material icon
- [x] Turn Last prices into Latest prices
- [x] You should be able to click on "Open Prices" to come back to the home page
- [ ] The user menu has a logout icon if you're not logged in
- [ ] You should be able to click on the map popup to select a store
- [x] Can we add the time on the Latest prices to make it more exciting
- [ ] Add a "Top price per kilo page"
- [ ] Define a target basket on which to focus efforts
- [x] Add a "Add a new price" to the product page
- [ ] https://github.com/openfoodfacts/open-prices-frontend/issues/318
- [ ] In Stats, for whimsical effect, add: Max price (for the whole db), Min Price (for the whole db), Number of cities with at least a price, Number of countries with at least a price, Number of different stores with at least a price.
- [ ] Add a timeline view on the product page for a given store (with a store selector, and an option to add a new pricepoint for that specific product and that specific store)
- [ ] Allow to store favorite stores
- [ ] Allow people to become the "Mayor" of stores
- [ ] Attempt to detect location (and store), timestamp from EXIF metadata, and to attempt extracting barcodes from image
- [ ] Allow to change language
- [ ] Allow to force language in the URL scheme
- [ ] Allow prepopulation of the form in the URL Scheme (barcode, store, currency, date)
- [ ] Use Wikidata to get a photo of products without barcodes using their category
- [ ] https://github.com/openfoodfacts/open-prices-frontend/issues/100
- [ ] Create a worldwide real-time map of prices being added
- [ ] Create a worldwide map of stores with at least a price
- [ ] https://github.com/openfoodfacts/open-prices-frontend/issues/274
- [ ] Add a shortcut to promote the Hunger Games quantity game: https://hunger.openfoodfacts.org/questions?type=product_weight
- [ ] https://github.com/openfoodfacts/open-prices-frontend/issues/447
- [ ] Add links to Top Products, Top Locations, Top Users from the Stats page
```

psankhe28:
@teolemon I'm interested in contributing to the implementation of a dark mode for this project. May I proceed with this task? Additionally, are we considering a specific color theme for the dark mode, or should I stick to a classic white and black scheme?

raphodn:
@psankhe28 I created the issue, go ahead ! maybe something classic/basic to start with, and we'll improve it in future iterations :)

[...]

PR #255, opened 2024-01-27
feat: Make the sign-in button similar to Open Food Facts

alexfauquette:
From #65 

> The user menu has a logout icon if you're not logged in

I think Pierre got mislead by the sign in  icon which is very close to the sign out. I propose To follow OFF behavior: if sign in display the name otherwise display "sign in"

image

This PR is mostly for me to get familiar back with vue.js If you prefer the previouse behavior no pb

### DIFF
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -6,14 +6,12 @@
       {{ APP_NAME }}
     </v-app-bar-title>
     <template v-slot:append>
-      <v-btn v-if="!username" to="/sign-in" icon="mdi-login"></v-btn>
+      <v-btn v-if="!username" to="/sign-in" prepend-icon="mdi-account" class="text-none">Sign in</v-btn>
       <v-menu v-if="username">
         <template v-slot:activator="{ props }">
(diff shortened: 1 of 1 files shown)


src/components/Header.vue
```diff
@@ -6,14 +6,12 @@
       {{ APP_NAME }}
     </v-app-bar-title>
     <template v-slot:append>
-      <v-btn v-if="!username" to="/sign-in" icon="mdi-login"></v-btn>
+      <v-btn v-if="!username" to="/sign-in" prepend-icon="mdi-account" class="text-none">Sign in</v-btn>
```
raphodn, line 9: missing translation

raphodn (commented):
maybe have a different behavior on small screens (smartphones ?)
I feel like the space will be tight... I'd rather keep the "Open Prices" title, and show only icons on small screens, with text on bigger screens ?

dq18:
I would also either have a different logo or the color of the logo if signed in or signed out

Issue #224, opened 2024-01-25
Translate raw product categories


raphodn:
### Story

Currently the list of categories is only limited to english : https://github.com/openfoodfacts/open-prices-frontend/blob/master/src/data/category-tags.json

See #30 on how it was initiated

the full json is here : https://static.openfoodfacts.org/data/taxonomies/categories.full.json

raphodn:
Currently the `categories-tags.json` is only a subset of all the categories. We could add the available translations in english.

@raphael0202 did the filtering. Could you share your script ? So that we can use it to add the other languages ?

raphael0202:
I built it manually by fetching children of "fruits", "vegetables" and selecting them.
If we use the autocomplete feature of search-a-licious, we won't be able to have this filtered set of categories.
One quicker way of doing this would be to add all translations to the JSON file. The full category taxonomy in JSON is not that big, so with a reduced set I think it's a reasonable approach.

raphodn:
@raphael0202 did you have a look at the script to filter the raw categories + translate them ? 
- script here : https://github.com/openfoodfacts/open-prices-frontend/blob/master/data/categories/filter_categories.py
- additional filtering or categories needed ?
- should this (script & jsons) be in the frontend, or in the backend ?

raphael0202:
It looks good to me!
Keeping it front-end is good to me as long as the bundle size is reasonable.

PR #261, opened 2024-01-30
feat: Category translation

dq18:
### What
- call Searchalicious API to get list of category in local language
- Will need to test on preprod server to check CORS issues
- WIP

### DIFF
--- a/src/views/AddPriceMultiple.vue
+++ b/src/views/AddPriceMultiple.vue
@@ -374,7 +374,7 @@ export default {
       ],
       productMode: null,
       category_query: null,
-      categoryTags: null,  // list of category tags for autocomplete - initialized in mounted with API call to a fixed value
+      categoryTags: [],  // list of category tags for autocomplete - initialized in mounted with API call to a fixed value
       category_loading: false,
       originsTags: OriginsTags,  // list of origins tags for autocomplete
       labelsTags: LabelsTags,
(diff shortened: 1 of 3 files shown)


src/views/AddPriceMultiple.vue
```diff
@@ -434,6 +439,41 @@ export default {
...
+      // get a list of categories to init the autocomplete - might be replaced with hardcoded categories
+      const initVal = 'a'
+      this.queryCategory(initVal)
+    },
+    queryCategory(newVal) {
+      this.loading = true
+      if (newVal && newVal.length > 0) {
+        this.category_loading = true
+        api
+          .getCategories(newVal, 'category', 10)
```
raphodn, line 451: on every new price(s) we fetch the categories ? this isn't ideal :sweat_smile: 
- we could fetch them when the language is changed in the user settings
- we could have them directly in the raw json

raphodn: maybe document some implementation solutions first in the corresponding issue ? #224

dq18: Here the API is called:
- once at page mount (with a simple query "a")
- Everytime the user inputs something in autocomplete field. So for example if the user type "rad", it will call for "r", then "ra" and "rad"

The fetch returns a list of max 10 categories (max number can be changed) and this list will be shown in the autocomplete field.

I will list solutions in the issue

raphodn:
Done in #273 & #275 & #276 instead

PR #266, opened 2024-01-30
feat: add category price per unit toggle

dq18:
### What

- Added a choice selector to enable changing `price_per` between `KILOGRAM ` or `UNIT` when inputting a price for a product without barcode
- Default behavior: automatically set to Kilogram

### Screenshot

image

### DIFF
--- a/src/views/AddPriceMultiple.vue
+++ b/src/views/AddPriceMultiple.vue
@@ -202,11 +202,14 @@
             <i>{{ $t('AddPriceSingle.ProductInfo.SetProduct') }}</i>
           </p>
           <h3 class="mb-1">
-            <i18n-t keypath="AddPriceSingle.PriceDetails.Text" tag="span">
-              <template #perKg>
-                <span v-if="productMode === 'category'">{{ $t('AddPriceSingle.PriceDetails.TextPerKg') }}</span>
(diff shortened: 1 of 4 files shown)


src/views/AddPriceMultiple.vue
raphodn, line 1: i see a few things missing : 
- when a user submits a price - `createPrice()` - with barcode, we should set the `price_per` to null
- we need to change the `PriceCard` as well

raphodn (commented):
awesome !

src/views/AddPriceMultiple.vue
raphodn: and i added a screenshot (helpful for reviewers and people stumbling on the PR !)

dq18: - that indeed happened in AddPriceMultiple. I added a cleanup in both AddPriceMultiple and AddPriceSingle for peace of mind
- PriceCard: Updated
image

raphodn: top !

raphodn: `Kg` -> `kg` and `Unit` -> `unit` no ?
appart from that it's mergable :)

PR #273, opened 2024-02-03
feat: script to filter and generate raw categories (with translations)

raphodn:
### What

Script to automatically :
- filter the taxonomy categories to keep only the "raw" ones (needed for the price add form + price display)
- generate the language specific fields

### Output

```
Total number of categories: 12854
Filter with the following parent categories: [<TaxonomyNode en:vegetables>, <TaxonomyNode en:fruits>, <TaxonomyNode en:culinary-plants>, <TaxonomyNode en:nuts>, <TaxonomyNode en:potatoes>, <TaxonomyNode en:textured-vegetable-protein>]
Add extra nodes: ['en:rolled-oats', 'en:ginger', 'en:mushrooms']
Additional filtering on: ['Cooked ', 'Fresh ']
Categories remaining: 579
Wrote to language files
```

185 json files in `src/data/categories/<language-code>.json`

### Links that helped

- https://github.com/openfoodfacts/openfoodfacts-python/blob/develop/openfoodfacts/taxonomy.py
-


dq18:
I would rather have this script in the backend and have a mechanism to fetch the needed data and store it in the store. But good for now!

dq18 (approved):

Issue #277, opened 2024-02-04
Mismatch between language list & crowdin locales


raphodn:
### Story

The file `i18n/data/languages.json` has a list of languages, but their code don't all correspond to the (official ?) crowdIn locale list (in `i18n/locales/...`).
This list of languages is also used to generate the category translations...

This list of languages was generated in #241 from https://github.com/annexare/Countries/blob/main/packages/countries/src/data/languages.ts

dq18:
@raphodn I can add a library that gets the user location (country) by IP and then automatically select the country.
However, I think it will be better to have an onboarding page, basically bringing up the Settings page the first time a user logs in the app. What do you think?

raphodn:
The issue here is first and foremost about improving the "language selector" in the user settings, to match it with the full list provided by crowdin.

I think a country selector is something different, and should be done in another issue..

dq18:
I edited the story you created to explain why I need to the country selector and tasks. Maybe you did not see the change.

Crowdin generates locales with format like `nl-NL.json` . If I improve the language selector right away it will disable a few languages like nl, pt, zh, sr.
For 'en', since it is our template default, it is not an issue, though `en-GB.json, en-AU.json` will never be used.
We have `nl.json` and 'pt.json' that were originally created by me but not updated by Crowdin and will need to be removed

raphodn:
Also, en / en-GB / en-AU actually correspond to different languages ^^

for example : the wikipedia language selector in the account settings
![image](https://github.com/openfoodfacts/open-prices-frontend/assets/7147385/f6998f87-438b-4f6d-907c-d30a81ab2d67)

raphodn:
Todo : script to auto-generate the languages.json
- use crowdin list as input
- find a full list of languages (Didier used https://github.com/annexare/Countries/blob/main/packages/countries/src/data/languages.ts)
- fields : code, name, native

PR #296, opened 2024-02-07
chore: locale cleanup

dq18:
### What

- Fix issue #277


src/i18n/data/languages.json
raphodn, line 1: how did you generate (or update) the `languages.json` ?

raphodn: can we document it in a README.md file next to it ? or in this PR ? thanks !

dq18: That was done first by comparing the locale list of Crowdin with the former language.json file and removing languages that are not in the locale list of Crowdin. Then manually add the locale of Crowdin which were not in language.json.
To get the names and native names, I used:
https://en.wikipedia.org/wiki/ISO_639:c
https://chronoplexsoftware.com/localisation/help/languagecodes.htm
And when not in those links, google search

src/i18n/data/languages.json
```diff
@@ -911,23 +707,32 @@
...
+    "code": "zh_HK",
+    "name": "Chinese Traditional (Hong Kong)",
+    "native": "中文(香港特別行政區)"
   },
   {
-    "code": "zh",
-    "name": "Chinese",
-    "native": "中文"
+    "code": "zh_TW",
+    "name": "Chinese Traditional",
```
raphodn, line 734:
```suggestion
    "name": "Chinese Traditional (Taiwan)",
```

src/i18n/data/languages.json
raphodn: ok ideally we should a script somewhere :)

raphodn (approved):

PR #297, opened 2024-02-07
feat: add a country selector in the user settings

dq18:
### What

In the user settings, add a country selector

### Why ?

- to later have currency auto-selection
- stats/analytics (?)

### Screenshot

image


src/i18n/data/countries.json
raphodn, line 1: how did you generate this `countries.json` ? can you document it in a seperate README.md ? (in case we need to update it, new countries can appear ^^)

src/i18n/data/countries.json
```diff
@@ -0,0 +1,1501 @@
...
+      "native": "Österreich",
+      "capital": "Vienna",
+      "currency": ["EUR"],
+      "languages": ["de"]},
+    {"code": "AU",
+      "name": "Australia",
+      "native": "Australia",
+      "capital": "Canberra",
+      "currency": ["AUD"],
+      "languages": ["en"]},
```
raphodn, line 79: `en_AU` instead ?

src/i18n/data/countries.json
raphodn: if there is no changes compared to this library, why not add it directly to the repo ?
`npm install countries-list`
^^

dq18: I removed some info that is not relevant from the countries list, like "Continent" and "Phone", and we will likely change the information in the future. The languages list through this library would also need a lot of workaround since locale codes are missing and it is not handling locales with country specificity like zh-TW

src/i18n/data/countries.json
dq18: We should keep it like this here. If we later want to use the country/language relationship in our app, I would add some logic to:
- check if a locale with country code exists
- if not then get the locale without country code

I am going to use this logic to display the country languages first in the language dropdown in a future PR

src/views/UserSettings.vue
```diff
@@ -104,10 +116,12 @@ export default {
     initUserSettingsForm() {
       this.userSettingsForm.currency = this.appStore.user.last_currency_used
       this.userSettingsForm.selectedLanguage = this.languageList.find(lang => lang.code === localeManager.guessDefaultLocale()) || this.languageList.find(lang => lang.code === 'en')
+      this.userSettingsForm.selectedCountry = countryData.find(country => country.code === this.appStore.user.country) || countryData.find(country => country.code === 'FR')    
     },
     async updateSettings() {
       await localeManager.changeLanguage(this.userSettingsForm.selectedLanguage.code)
       this.appStore.setLanguage(this.userSettingsForm.selectedLanguage.code)
+      this.appStore.setCountry(this.userSettingsForm.selectedCountry)
```
raphodn, line 124:
```suggestion
      this.appStore.setCountry(this.userSettingsForm.selectedCountry.code)
```

src/views/UserSettings.vue
```diff
@@ -9,6 +9,15 @@
         <v-card :title="$t('UserSettings.ChangeLanguage')" prepend-icon="mdi-earth">
           <v-divider></v-divider>
           <v-card-text>
+            <v-autocomplete
+              v-model="userSettingsForm.selectedCountry"
+              label="Country"
```
raphodn, line 14: translation missing :)

src/store.js
```diff
@@ -10,6 +10,8 @@ export const useAppStore = defineStore('app', {
       last_currency_used: 'EUR',  // TODO: init with user locale ?
       recent_locations: [],
       language: localStorage.getItem('user-locale') || import.meta.env.VITE_DEFAULT_LOCALE,  // 'en'
+      country: null, // will be of type Object like
```
raphodn, line 13: shouldn't we init with 'FR' (instead of having a `||` in UserSettings) ?

src/views/UserSettings.vue
```diff
@@ -9,6 +9,15 @@
...
+              :items="countryList"
+              item-title="name"
```
raphodn, line 16:
```suggestion
              item-title="native"
```
like the languages ?

src/views/UserSettings.vue
dq18: Thanks for seeing this. Corrected. Translation key is already in locale files (from a previous PR)

src/store.js
dq18: Definitely, corrected. 'FR' is initialized in store

src/views/UserSettings.vue
dq18: Corrected

src/i18n/data/countries.json
raphodn: ok we'll still need a script to generate/update this country list somehow (like in #338 )

raphodn: https://github.com/mledoze/countries/tree/master/dist

PR #310, opened 2024-02-09
feat: Settings country dependencies

dq18:
### What
Thanks to #297, it is now possible to select a country.

- The Settings page has been updated to change the currency to the country currency automatically (it can still be changed by the user)
- The Settings page will also list the current language and the country languages at the top of the dropdown list
- Minor fixes


src/views/UserSettings.vue
```diff
@@ -94,7 +94,34 @@ export default {
         this.languageTranslationCompletion = await localeManager.calculateTranslationCompletion(this.userSettingsForm.selectedLanguage.code)
       }
     },
-
+    'userSettingsForm.selectedCountry': function (newValue, oldValue) {
```
raphodn, line 97: i just tested. it works, but it feels kinda strange that when changing the country, both the language & currency are changed without the user realizing :thinking: 
- we should have a short message popup (like when uploading a proof)
- or separate the behavior and not automatically make the changes..

src/views/UserSettings.vue
```diff
@@ -94,7 +94,34 @@ export default {
...
+      if (newValue !== oldValue) {
+        const selectedCountry = this.countryList.find(country => country.code === newValue)
+        // Update the currency list to the first currency of the selected country if it exists or USD otherwise
+        if (selectedCountry && selectedCountry.currency && selectedCountry.currency.length > 0) {
+          this.userSettingsForm.currency = selectedCountry.currency[0]
+        } else {
+          this.userSettingsForm.currency = 'USD'
```
raphodn, line 104: why is USD hardcoded here ? we have EUR set as default in the store. we could move everything to an env btw and have a `VITE_DEFAULT_CURRENCY` (like LOCALE & COUNTRY)

raphodn (approved):

src/views/UserSettings.vue
dq18: Language will not be automatically changed, the current selected language (or browser language or default English) will still be the one selected. But in the dropdown it will show up first, followed by languages of the country and then other languages.
I suppose the same can be done for currency. Or have a popup. But ultimately we will change the currency dropdown to a checkbox list for the user to choose currencies that he is paying with (issue #359 )
I will leave it that way for now and do another PR for currency checkbox dropdown.

src/views/UserSettings.vue
dq18: I think I used that for testing and forgot to remove. The default should revert to VITE_DEFAULT_CURRENCY set in the store and the "else" statement is useless here

dq18: It wasn't for testing. I changed to the store value

src/views/UserSettings.vue
raphodn: I tested again and the currency changes. can we have a small `snackbar` to let the user know ?

also the code below seems very complex... I'm not sure we need to go into all these automation, vs just let users choose a country / a language / a currency :thinking:

dq18: Alright, I will update #372 to take this into consideration and revert to decoupling of currency.

Issue #298, opened 2024-02-07
Bring the  Header background color in line with the other OFF projects


teolemon:
### Problem
Can we change the Header background color to rgb(242, 233, 228) to bring it in line with the other OFF projects ?

### Part of
- #220

Apurv428:
Hi @teolemon , can I take up this issue?

raphodn:
@Apurv428 no need to ask, go for it :)

raphodn:
1. in https://world.openfoodfacts.org/ the color seems to be `#f2e9e4` ?
2. same for the footer ?

PR #316, opened 2024-02-10
feat: header background color (same as OFF)

Apurv428:
### What
The header's background color has been updated to RGB(242, 233, 228) to align it with the color scheme used in other OFF projects.

### Screenshot

Before the changes:

image

After the header background color has been updated:

image

### Part of 
#298

### DIFF
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -1,5 +1,5 @@
 <template>
-    <v-app-bar :elevation="1" style="background-color: rgb(242, 233, 228);">
+  <v-app-bar :elevation="1" style="background-color: rgb(242, 233, 228);">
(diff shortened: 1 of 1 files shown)


src/components/Header.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-app-bar :elevation="1">
+    <v-app-bar :elevation="1" style="background-color: rgb(242, 233, 228);">
```
raphodn, line 2:
```suggestion
  <v-app-bar :elevation="1" style="background-color: rgb(242, 233, 228);">
```

raphodn (approved):
thanks :+1:

PR #320, opened 2024-02-10
feat: read gallery image exif date (and update if not today)

raphodn:
### What

- new JS library [ExifReader](https://github.com/mattiasw/ExifReader) to get the gallery image's `DateTimeOriginal`
- parse the date
- if different from today, update the price create form date (single & multiple mode)


teolemon (approved):
Really nice 🥳

Issue #336, opened 2024-02-14
Cleanup locale setup


raphodn:
### Story

Currently there are some issues in the locale init & management
- the user local is run & set on each page change
- the localstorage is used instead of the vue store
- we read the params but we don't need it for now

TTalex:
Closing, should be fixed with linked PR.

PR #337, opened 2024-02-15
chore: cleanup language settings and language selector

dq18:
### What
- fix issue #336


src/store.js
```diff
@@ -9,7 +9,6 @@ export const useAppStore = defineStore('app', {
       last_product_mode_used: 'barcode',
       last_currency_used: 'EUR',  // TODO: init with user locale ?
       recent_locations: [],
-      language: localStorage.getItem('user-locale') || import.meta.env.VITE_DEFAULT_LOCALE,  // 'en'
```
raphodn, line 12: we should keep the `user.language` key here + a default value (null or VITE_DEFAUL_LOCALE) + indicate it will be a code (ex : 'en')

src/views/UserSettings.vue
```diff
@@ -113,12 +113,12 @@ export default {
   methods: {
     initUserSettingsForm() {
       this.userSettingsForm.currency = this.appStore.user.last_currency_used
-      this.userSettingsForm.selectedLanguage = this.languageList.find(lang => lang.code === localeManager.guessDefaultLocale()) || this.languageList.find(lang => lang.code === 'en')
+      this.userSettingsForm.selectedLanguage = this.languageList.find(lang => lang.code === localeManager.guessDefaultLocale()).code
```
raphodn, line 116: it feels strange that the `guessDefaultLocale` happens in the settings form (vs on app startup).

here it means the new user arrived on the app, the app will show up by default as 'en', right ? even if the user has a 'French' browser ? And after logging in, and going to its settings page, then the form field language will show 'French' :thinking: 

Or I'm missing something ? :sweat_smile:

src/store.js
dq18: It should indeed be here. I messed up when merging the changes from master in 2nd commit and deleted the line.

raphodn (approved):

Issue #274, opened 2024-02-03
Add a dark mode


raphodn:
### Story

Add a toggle to switch the UI to dark mode
- remember the user choice
- depending on browser/device settings ?

psankhe28:
Apologies, @raphodn, I accidentally force-pushed. Additionally, I'm sorry for the delayed contribution to this pull request due to academic commitments. I have a question regarding the toggle button, as pointed out by teolomon. It's supposed to be in the settings, but I'm having trouble locating it. Can you provide some guidance on this matter?

I'm uncertain if you'll receive a notification or be able to view the comment on a closed pull request, so I'm reaching out here.

raphodn:
Vuetify has a "dark" theme : https://vuetifyjs.com/en/features/theme/#setup

raphodn:
A possible better version would be to select between 3 options: Light, Dark and System.

Example in https://vuetifyjs.com/en/ : 
![image](https://github.com/user-attachments/assets/e3459f55-6274-412b-940c-cc4252503cae)

PR #339, opened 2024-02-16
feat: Dark mode implemented

psankhe28:
What
Implemented a dark mode in the user settings file.

Screenshot
Dark mode:

image

Light mode:

image

System mode:

image

Part of
https://github.com/openfoodfacts/open-prices-frontend/issues/274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [1.43.0](https://github.com/openfoodfacts/open-prices-frontend/compare/v1.42.0...v1.43.0) (2024-02-20)
+
+
+### Features
+
+* add price filter ([#358](https://github.com/openfoodfacts/open-prices-frontend/issues/358)) ([1d5163d](https://github.com/openfoodfacts/open-prices-frontend/commit/1d5163d8d617f173aae6378ac3cbfbfb3cd10833))
+
+
+### Technical
+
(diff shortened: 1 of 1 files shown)


dq18:
Hi @psankhe28 , thanks for contributing.
It's a good start, there are a few things to correct.
- There is an issue with the footer that I can see on Chrome on mobile device mode and normal mode:
image
- There is no need for a "Follow System Settings" option

And in general, I would push the work further to have a uniform coloring, not just the background but also cards/buttons and choose the right font colors. Right now there are too many discrepancies for every page
image

I am definitely not a CSS/design expert, but don't hesitate to check other website for inspiration!

psankhe28:
I will fix the footer issue and remove the default system setting option too. If the background is fine, I'll modify cards, buttons, and other components. Please let me know. Thanks!

src/components/Footer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-footer class="bg-grey-lighten-1" style="max-height: 20%">
+  <v-footer class="bg-grey-lighten-1" style="position: absolute; bottom: -0; width: 100%;overflow: hidden;">
```
Valimp, line 2: Is the value "-0" desired?

Apurv428: It was supposed to be zero so that the footer sticks to the bottom; I mistakenly included the minus sign.
I will rectify it in the next push. 
Sorry for that!

PR #342, opened 2024-02-16
fix(l10n): improve language script

dq18:
### What

Following #338 

- Added/corrected some language data
- Better visual file output


raphodn (approved):
:+1:

PR #344, opened 2024-02-16
feat: update URL when usering filter & order_by queries

dq18:
### What

- save filter choices in url query parameters
- BrandDetail and ProductList

```
http://localhost:5173/products?filter=&order_by=-unique_scans_n
http://localhost:5173/products?filter=&order_by=-price_count
http://localhost:5173/products?filter=hide_price_count_gte_1&order_by=-price_count
```


src/views/BrandDetail.vue
```diff
@@ -130,11 +130,13 @@ export default {
     },
     toggleProductFilter(filterKey) {
       this.productFilter = this.productFilter ? '' : filterKey
+      this.$router.push({ query: { ...this.$route.query, productFilter: this.productFilter } })
```
raphodn, line 133: I did a test `http://localhost:5173/products?productOrder=-price_count&productFilter=hide_price_count_gte_1`
- I would rename `productOrder` to `order_by` to match with the backend call
- for `productFilter` it's trickier, at least rename to `filter`, though the API call is simply `price_count=0`

raphodn: also, when I refresh the page, the URL is still the same, but the filters are NOT taken into account :/

raphodn: thanks for the fix ! actually it was calling `initBrand()` twice, so I added a commit - https://github.com/openfoodfacts/open-prices-frontend/pull/344/commits/b54f5f3368eabdc7b65e4dafc8e37b75a81c9e8a
- to fix this
- to move some stuff to the constants

raphodn: did the same with ProductList (added a watch on `$route`)

PR #345, opened 2024-02-16
fix(l10n): fix languages with multiple entries in name and native

dq18:
### What
- Some languages had several comma or semicolon separated names or native names, fixed script to take the first one
- Capitalized first letter

### DIFF
--- a/data/languages/generate_languages.py
+++ b/data/languages/generate_languages.py
@@ -64,8 +64,10 @@ def get_language_source():
         for key in row:
             # ld[key.strip()] = row[key].strip()
             ld["code"] = row["639-1 "].strip()
-            ld["name"] = row["Language name "].strip()
-            ld["native"] = row["Native name "].strip().strip("\u200e")  # remove left-to-right
+            name = row["Language name "].strip().split(',')[0].split(';')[0] # keep only the first language name if there is a , or ;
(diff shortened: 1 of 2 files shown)


data/languages/generate_languages.py
```diff
@@ -64,8 +64,10 @@ def get_language_source():
         for key in row:
             # ld[key.strip()] = row[key].strip()
             ld["code"] = row["639-1 "].strip()
-            ld["name"] = row["Language name "].strip()
-            ld["native"] = row["Native name "].strip().strip("\u200e")  # remove left-to-right
+            name = row["Language name "].strip().split(',')[0].split(';')[0] # keep only the first language name if there is a , or ;
+            ld["name"] = name[0].upper() + name[1:] # capitalize first letter and keep the rest as it is (capitalize function lower the rest of the string)
```
raphodn, line 68: python has a [capitalize()](https://docs.python.org/3/library/stdtypes.html#str.capitalize) function

raphodn (approved):
:+1:

data/languages/generate_languages.py
dq18: I tried `capitalize()` at first but it lowers the rest of the string (I wrote it in comments on line 68).
String like ""Bahasa Melayu" would then be converted to ""Bahasa melayu". Another idea is to split, capitalize and concat again.
Anyway the source is not uniform and would require manual editing

raphodn: ok perfect then :)

PR #346, opened 2024-02-17
fix: show tooltip on click (discount, date)

dq18:
### What
- added possibility to show on click tooltips for "discount" and "date" for price cards since there is hovering is not feasible on touchscreen


raphodn (approved):
cool !

Issue #32, opened 2023-12-19
Favicon, Metatags


raphodn:
### Story

- we need a favicon
- we need to add Metatags - https://metatags.io/

raphodn:
@teolemon do you have a logo or favicon in mind ?
Or should we use the Open Food Facts one for now ? https://world.openfoodfacts.org/images/favicon/favicon.ico

raphodn:
@teolemon interested in the OFF greyed icon set :)

teolemon:
https://github.com/openfoodfacts/openfoodfacts-server/blob/5b330154c42282efc72b0a756a49bc455dac0f7a/html/images/logos/logo-variants/MONO_RVB_ICON_WHITE_BG_OFF.svg?plain=1
https://github.com/openfoodfacts/openfoodfacts-server/blob/5b330154c42282efc72b0a756a49bc455dac0f7a/html/images/logos/logo-variants/B_W_CMJN_ICON_WHITE_BG_OFF.svg
https://github.com/openfoodfacts/openfoodfacts-server/blob/5b330154c42282efc72b0a756a49bc455dac0f7a/html/images/logos/logo-variants/MONO-B_W_CMJN_ICON_BLACK_BG_OFF.svg
https://github.com/openfoodfacts/openfoodfacts-server/blob/5b330154c42282efc72b0a756a49bc455dac0f7a/html/images/logos/logo-variants/MONO-B_W_CMJN_ICON_WHITE_BG_OFF.svg
https://github.com/openfoodfacts/openfoodfacts-server/blob/5b330154c42282efc72b0a756a49bc455dac0f7a/html/images/logos/logo-variants/MONO_RVB_ICON_BLACK_BG_OFF.svg

raphodn:
thanks ! I took : 
- favicon : https://github.com/openfoodfacts/openfoodfacts-server/blob/5b330154c42282efc72b0a756a49bc455dac0f7a/html/images/logos/logo-variants/B_W_CMJN_ICON_WHITE_BG_OFF.svg
- social image : https://github.com/openfoodfacts/openfoodfacts-server/blob/main/html/images/logos/logo-variants/B_W_CMJN_VERTICAL_WHITE_BG_OFF.svg

PR #348, opened 2024-02-17
fix: fix Lighthouse errors (PWA config)

ArturLange:
### What
- Added standard PWA compatible icons that were missing.
- Fixed some other errors from Google Lighthouse (accessibility etc.)
- Changed open graph image to absolute path as FB does not recognise relative image - see below
Also see #32 

image


src/components/Header.vue
```diff
@@ -2,21 +2,21 @@
...
         <template v-slot:activator="{ props }">
           <v-btn v-bind="props" icon="mdi-account-circle"></v-btn>
         </template>
         <v-list>
           <v-list-item :slim="true" prepend-icon="mdi-account" disabled>{{ username }}</v-list-item>
           <v-divider></v-divider>
-          <v-list-item :slim="true" prepend-icon="mdi-view-dashboard-outline" to="/dashboard">{{ $t('Header.Dashboard') }}</v-list-item>
-          <v-list-item :slim="true" prepend-icon="mdi-cog-outline" to="/settings">{{ $t('Header.Settings') }}</v-list-item>
-          <v-list-item :slim="true" prepend-icon="mdi-logout" @click="signOut">{{ $t('Header.Sign-out') }}</v-list-item>
+          <v-list-item aria-label="Dashboard" :slim="true" prepend-icon="mdi-view-dashboard-outline" to="/dashboard">{{ $t('Header.Dashboard') }}</v-list-item>
```
raphodn, line 17: we should use the translations here

.gitattributes
raphodn, line 1: can you explain the purpose ?

raphodn (approved):
thanks !

.gitattributes
ArturLange: I added image files to Git LFS.
AFAIK adding them to Git repository normally can bloat the repository over time, this should be safer

raphodn: it seems to be an [addon](https://stackoverflow.com/questions/71745178/gitlfs-images-and-or-files-how-do-i-view-them-how-do-i-download-them), and only for git, and only for a couple of images here. I don't think we need this ;)

ArturLange: You're right, it might not be worth it :)

teolemon (approved):
Swell

public/android-chrome-192x192.png
raphodn, line 1: I have an error in my console when running this branch locally

image

raphodn: and github says 

image

ArturLange: Good catch, they were messed up by adding and removing LFS :crying_cat_face: 
I added them again, they should work now.

PR #349, opened 2024-02-17
perf: Dynamic imports of vue components

ArturLange:
### What

Changed components loading to dynamic. That way those can be split into chunks by `vite` and `rollup` and we can improve performance by loading smaller bundles.

Linked to #88

PR #350, opened 2024-02-17
build: upgrade Vue

ArturLange:
### What

Upgrade `vue` for dynamic import of components to work in #349 

v3.3.9 --> v3.4.19

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1171,7 +1171,7 @@ vue@^3.2.25:
     "@vue/server-renderer" "3.3.13"
     "@vue/shared" "3.3.13"
 
-vue@^3.3.9:
+vue@^3.4.19:
   version "3.4.19"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1123,15 +1172,15 @@ vue@^3.2.25:
     "@vue/shared" "3.3.13"
 
 vue@^3.3.9:
```
raphodn, line 1174: which command did you run ? `yarn upgrade` right ?

maybe change the the `package.json` as well to reflect the change ? `^3.4.19`

https://github.com/vuejs/core/blob/main/CHANGELOG.md

ArturLange: I used `yarn upgradeInteractive`. 
`3.4.19` fulfills `^3.3.9` so it should be valid.

raphodn: Yet the version change would just be cosmetic ^^

ArturLange: Done then :slightly_smiling_face:

raphodn (approved):
:+1:

PR #352, opened 2024-02-17
fix: Turn 2 links into real buttons (Type a barcode & Select from gallery)

### LINKED ISSUE #329: Turn Type a barcode and Select from gallery into real buttons

ArturLange:
### What

Following #164 & #162

Change **Type a barcode** and **Select from gallery** links to buttons

### Screenshot

||Image|
|--|---|
|Small screens|image|
|Larger screens|image|

### DIFF
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -29,8 +29,14 @@
               </v-item-group>
             </h3>
             <v-sheet v-if="productMode === 'barcode'">
-              <v-btn class="mb-2 mr-2" size="small" prepend-icon="mdi-barcode-scan" @click="showBarcodeScanner">{{ $t('AddPriceSingle.ProductInfo.ScanBarcode') }}</v-btn>
-              <v-btn class="mb-2 mr-2" size="small" prepend-icon="mdi-numeric" @click.prevent="showBarcodeManualInput">{{ $t('AddPriceSingle.ProductInfo.TypeBarcode') }}</v-btn>
+              <v-btn class="mb-2 mr-2" size="small" prepend-icon="mdi-barcode-scan" @click="showBarcodeScanner">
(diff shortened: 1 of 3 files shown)


src/views/AddPriceSingle.vue
```diff
@@ -30,7 +30,7 @@
             </h3>
             <v-sheet v-if="productMode === 'barcode'">
               <v-btn class="mb-2 mr-2" size="small" prepend-icon="mdi-barcode-scan" @click="showBarcodeScanner">{{ $t('AddPriceSingle.ProductInfo.ScanBarcode') }}</v-btn>
-              <a href="#" @click.prevent="showBarcodeManualInput">{{ $t('AddPriceSingle.ProductInfo.TypeBarcode') }}</a>
+              <v-btn class="mb-2 mr-2" size="small" prepend-icon="mdi-numeric" @click.prevent="showBarcodeManualInput">{{ $t('AddPriceSingle.ProductInfo.TypeBarcode') }}</v-btn>
```
raphodn, line 33: I'm wondering if we should have a different behavior on small screens : a shortened version to avoid having 2 lines. Can be done in a different PR of course

image

raphodn (approved):
:100:

src/views/AddPriceSingle.vue
dq18: Was about to work on this! But no worries, great that you are helping @ArturLange , greatly needed.
I would shorten also to "Scan" and "Manual" (or similar). I was thinking of just using obvious icons too, like a "camera" icon and a "keypad" icon. Some UI/UX expert or designer input would be great

raphodn: The UX "good practice" (at least used at OFF) is to have always icon + text.

I'm adding a commit (and screenshot) to reduce the text on small screens.

PR #355, opened 2024-02-17
fix: typos in dynamic import (ref #349)

dq18:
### What
- fix dynamic import for page brandDetail following PR #349


raphodn (approved):
thanks :sweat_smile:

PR #367, opened 2024-02-21
feat(products): display categories & labels count

raphodn:
### What

Thanks to the new fields in the backend, we can start displaying a badge with the count
- `Product.categories_tags` : https://github.com/openfoodfacts/open-prices/issues/154
- `Product.labels_tags` : https://github.com/openfoodfacts/open-prices/issues/216

### Screenshot

|Before|After|
|---|---|
|image|image|


dq18 (approved):
Everything is fine, we will need a chip cleanup / reorder / redesign eventually

raphodn:
Thanks for the review, I'll merge once we actually have data in these fields.
Currently they seem empty... 
Example : https://prices.openfoodfacts.org/api/v1/products?page=1&size=10&order_by=-price_count

PR #369, opened 2024-02-21
feat(products): display popup with product categories

raphodn:
### What

Following #367

When the user clicks on the new "product categories count chip", a popup opens with the list of the categories.

### Screenshot

|Product without categories|Product with categories|
|---|---|
|image|image|

### DIFF
--- a/src/components/ProductCategoriesDialog.vue
+++ b/src/components/ProductCategoriesDialog.vue
@@ -8,7 +8,7 @@
       <v-divider></v-divider>
 
       <v-card-text v-if="categories.length">
-        <v-chip v-for="category in categories" label class="mr-2 mb-2">
+        <v-chip v-for="category in categories" :key="category" label class="mr-2 mb-2">
           {{ category }}
         </v-chip>
       </v-card-text>
--- a/src/components/ProductCategoriesDialog.vue
+++ b/src/components/ProductCategoriesDialog.vue
@@ -1,0 +1,41 @@
+<template>
+  <v-dialog>
+    <v-card>
+      <v-card-title>
+        {{ $t('ProductCategories.Title') }} <v-btn style="float:right;" variant="text" density="compact" icon="mdi-close" @click="close"></v-btn>
+      </v-card-title>
+
+      <v-divider></v-divider>
+
+      <v-card-text v-if="categories.length">
+        <v-chip v-for="category in categories" label class="mr-2 mb-2">
+          {{ category }}
+        </v-chip>
+      </v-card-text>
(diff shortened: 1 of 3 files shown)


src/components/ProductCategoriesDialog.vue
```diff
@@ -0,0 +1,41 @@
...
+    <v-card>
+      <v-card-title>
+        {{ $t('ProductCategories.Title') }} <v-btn style="float:right;" variant="text" density="compact" icon="mdi-close" @click="close"></v-btn>
+      </v-card-title>
+
+      <v-divider></v-divider>
+
+      <v-card-text v-if="categories.length">
+        <v-chip v-for="category in categories" :key="category" label class="mr-2 mb-2">
+          {{ category }}
```
dq18, line 12: I would use `category.name` here so that non-english speaker can use the information.

And eventually in the future have a small symbol next to untranslated names with a tooltip saying that category is not translated with a link to taxonomies.
Also a link to add categories for this product.

src/components/ProductCategoriesDialog.vue
```diff
@@ -0,0 +1,41 @@
...
+        </v-chip>
+      </v-card-text>
+      <v-card-text v-if="!categories.length">
+        <span class="text-red">{{ $t('ProductCategories.Empty') }}</span>
```
dq18, line 16: Link to the OFF product page maybe?

src/i18n/locales/en.json
```diff
@@ -182,6 +182,10 @@
 		"ProductQuantityLitre": "{0} L",
 		"UnknownProduct": "Unknown product name"
 	},
+	"ProductCategories": {
+		"Title": "Categories",
+		"Empty": "This product does not have any categories yet."
```
dq18, line 187: Maybe:
"No categories have been added for this product yet." Or with extra "Help us @OFF page"

dq18 (approved):

src/components/ProductCategoriesDialog.vue
raphodn: the `categories` object looks like this : `["en:breakfasts", "en:spreads"]` . so the `category` is only a string

dq18: Alright!

src/components/ProductCategoriesDialog.vue
raphodn: yeah we should ! will add in another PR (currently the `ProductCategoriesDialog` component doesn't have any info about the product, need to change it)

PR #373, opened 2024-02-23
fix: force numeric keyboard for barcode & price inputs

raphodn:
### What

There was already `type="number"` (following https://github.com/openfoodfacts/open-prices-frontend/commit/e547914f057fb979c49e21daa7f94c673f1db6a0)
- added `imputmode="numeric"` for barcode inputs
- added `imputmode="decimal"` for price inputs

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

If not enough for some devices, we can also add a `pattern` param (next step)


ArturLange (commented):
Typo - should be `inputmode` and not `imputmode`.
Otherwise looks good

raphodn:
> Typo - should be `inputmode` and not `imputmode` everywhere. Otherwise looks good

oops, nice catch, thanks !

PR #375, opened 2024-02-24
feat: new proof detail page

raphodn:
### What

Thanks to https://github.com/openfoodfacts/open-prices/pull/228 & following #383

Allow users to view more details about their proofs, in a dedicated page
- the proof image & basic info
- the list of prices associated

Link is done via the proof price count chip.

An error message is shown is the proof is not found or not the owner

### Why

See prices filtered by a given proof, allow in the future to add a new price for this proof

### Screenshot

|Existing proof|Unknown proof|
|---|---|
|image|image|

### Todo (future PRs)

- alternative proof card (horizontal) to better display the info ?
- improve proof footer
- link to price create form (thanks to #376)
- manage price deletion "refresh" (list, count)

### DIFF
--- a/src/components/PriceCard.vue
+++ b/src/components/PriceCard.vue
@@ -35,7 +35,7 @@
         </v-col>
       </v-row>
 
-      <PriceFooter v-if="price && !hidePriceFooter" class="mt-2" :price="price" :hidePriceLocation="hidePriceLocation" :readonly="readonly"></PriceFooter>
+      <PriceFooter v-if="price && !hidePriceFooter" class="mt-2" :price="price" :hidePriceLocation="hidePriceLocation" :hidePriceProof="hidePriceProof" :readonly="readonly"></PriceFooter>
     </v-container>
   </v-card>
 </template>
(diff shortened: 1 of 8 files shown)


dq18:
`proofPriceList` of `ProofDetail` is not updated at price removal unless a new API call is made

raphodn:
> `proofPriceList` of `ProofDetail` is not updated at price removal unless a new API call is made

I tested and the price div is effectively removed, though indeed there is no action on the `proofPriceList` : same behavior as all other price lists, until we implement an equivalent of #328 ? But that will be addressed in another PR, no ?

ArturLange (approved):
LGTM

dq18:
> > `proofPriceList` of `ProofDetail` is not updated at price removal unless a new API call is made
> 
> I tested and the price div is effectively removed, though indeed there is no action on the `proofPriceList` : same behavior as all other price lists, until we implement an equivalent of #328 ? But that will be addressed in another PR, no ?

I think you mean the store system in #376 ? And yes, it will be addressed in another PR

PR #376, opened 2024-02-24
feat(price create form): allow user to select a recent proof

dq18:
### What

Allow user to select an existing proof when adding a price.

This is a WIP and only implemented for addSinglePrice. Please add your feedback to this PR on design and code and I will change / implement further.

### How
A new button allows to open a dialog with the last (max 10) proofs.
image
They will be displayed in a 5x2 grid.
image
Selecting one will highlight it and show a confirmation dialog
image
If "No" is selected, user goes back to choosing the proof.
If "Yes", it closes both dialogs and displays the selected proof as is done for "Camera" or "Gallery" modes
image

<hr>

PR simplified, no more confirmation dialog, works for both single & multiple price forms, rename "existing" to "recent"

### DIFF
--- a/src/components/ExistingProofDialog.vue
+++ b/src/components/ExistingProofDialog.vue
@@ -1,14 +1,14 @@
 <template>
-  <v-dialog>
+  <v-dialog max-height="80%" width="80%">
     <v-card>
       <v-card-title>
         {{ $t('ExistingProofDialog.ChooseLatestProofs') }} <v-btn style="float:right;" variant="text" density="compact" icon="mdi-close" @click="close"></v-btn>
       </v-card-title>
       <v-divider></v-divider>
-      <v-card-text>
+      <v-card-text style="overflow-y:scroll;">
         <v-row>
           <v-col cols="12" sm="6" md="3" v-for="proof in userProofList" :key="proof">
-            <ProofCard :proof="proof" :hideProofDelete="true" :isSelectable="true" :isSelected="selectedProof === proof" @proofSelected="selectProof"></ProofCard>
+            <ProofCard :proof="proof" :hideProofDelete="true" :isSelectable="true" @proofSelected="selectProof"></ProofCard>
           </v-col>
         </v-row>
     </v-card-text>
--- a/src/components/ExistingProofDialog.vue
+++ b/src/components/ExistingProofDialog.vue
@@ -7,25 +7,12 @@
       <v-divider></v-divider>
       <v-card-text>
         <v-row>
-          <v-col cols="6" sm="6" md="3" v-for="proof in userProofList" :key="proof">
-            <ProofCard :proof="proof" :reducedProofFooter="true" :isSelectable="true" :isSelected="selectedProof === proof" @proofSelected="selectProof"></ProofCard>
+          <v-col cols="12" sm="6" md="3" v-for="proof in userProofList" :key="proof">
+            <ProofCard :proof="proof" :hideProofDelete="true" :isSelectable="true" :isSelected="selectedProof === proof" @proofSelected="selectProof"></ProofCard>
           </v-col>
         </v-row>
     </v-card-text>
     </v-card>
-    <v-bottom-sheet v-model="showConfirmationPopup">
-      <v-card>
-        <v-card-text class="d-flex align-center">
-          {{ $t('ExistingProofDialog.SelectThis') }}
-          <v-btn class="mb-2 ml-2" size="small" prepend-icon="mdi-check-circle" @click="confirmSelection">
-            <span class="d-sm-inline-flex">{{ $t('Common.Yes') }}</span>
-          </v-btn>
-          <v-btn class="mb-2 ml-2" size="small" prepend-icon="mdi-close-circle-outline" @click="cancelSelection">
-            <span class="d-sm-inline-flex">{{ $t('Common.No') }}</span>
-          </v-btn>
-        </v-card-text>
-      </v-card>
-    </v-bottom-sheet>
   </v-dialog>
 </template>
 
@@ -46,7 +33,6 @@ export default {
       userProofPage: 1,
       loading: false,
       selectedProof: null,
-      showConfirmationPopup: false,
     }
   },
   computed: {
@@ -74,16 +60,8 @@ export default {
     },
     selectProof(proof) {
       this.selectedProof = proof
-      this.showConfirmationPopup = true
-    },
-    confirmSelection() {
-      this.showConfirmationPopup = false
-      this.$emit('proofConfirmed', this.selectedProof);
+      this.$emit('proofConfirmed', this.selectedProof)
       this.close()
-    },
-    cancelSelection() {
-      this.selectedProof = null;
-      this.showConfirmationPopup = false
     },
     close() {
       this.$emit('close')
@@ -91,9 +69,3 @@ export default {
   }
 }
 </script>
-
-<style scoped>
-.d-flex {
-  display: flex;
-}
-</style>
--- a/src/components/ExistingProofDialog.vue
+++ b/src/components/ExistingProofDialog.vue
@@ -1,0 +1,102 @@
+<template>
+  <v-dialog>
+    <v-card>
+      <v-card-title>
+        {{ $t('UserDashboard.LatestProofs') }} <v-btn style="float:right;" variant="text" density="compact" icon="mdi-close" @click="close"></v-btn>
+      </v-card-title>
+      <v-divider></v-divider>
+      <v-card-text>
+        <v-row>
+          <v-col cols="6" sm="6" md="3" v-for="proof in userProofList" :key="proof">
+            <ProofCard :proof="proof" :showProofFooter="false" :isSelectable="true" :isSelected="selectedProof === proof" @proofSelected="selectProof"></ProofCard>
+          </v-col>
+        </v-row>
+    </v-card-text>
+    </v-card>
+    <v-dialog v-model="showConfirmationPopup">
+      <v-card>
+        <v-card-title>Confirmation</v-card-title>
+        <v-card-text>Are you sure you want to select this proof?</v-card-text>
+        <v-card-actions>
+          <v-btn text @click="confirmSelection">Yes</v-btn>
+          <v-btn text @click="cancelSelection">No</v-btn>
+        </v-card-actions>
+      </v-card>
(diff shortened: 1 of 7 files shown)


teolemon:
@dq18 a number of things, more or less necessary, come to mind:
- is it possible to zoom on the proofs before selecting them ? 
- Adding a notion of time in the list of proofs might prove useful as well edge to edge line (Today, yesterday, etc ?)
- does it reselect the time of the proof in the price addition form ?
- should we allow linking proofs to places "asynchronously" (from your personal proof list, when you have free time), and thus prepopulate the place as well when selecting an existing proof ?

src/components/ExistingProofDialog.vue
```diff
@@ -0,0 +1,102 @@
...
+  <v-dialog>
+    <v-card>
+      <v-card-title>
+        {{ $t('UserDashboard.LatestProofs') }} <v-btn style="float:right;" variant="text" density="compact" icon="mdi-close" @click="close"></v-btn>
+      </v-card-title>
+      <v-divider></v-divider>
+      <v-card-text>
+        <v-row>
+          <v-col cols="6" sm="6" md="3" v-for="proof in userProofList" :key="proof">
+            <ProofCard :proof="proof" :showProofFooter="false" :isSelectable="true" :isSelected="selectedProof === proof" @proofSelected="selectProof"></ProofCard>
```
raphodn, line 11: i was thinking of doing a horizontal version of the ProofCard, but can be done after this PR & #375

raphodn: but like toelemon mentions, maybe showing part or all of the footer could still be useful, to have context on each proof

raphodn:
Looks great ! Could almost be deployed right now :)
- don't hide the proof footer to have context ?
- "preuve téléversée" is not accurate (but non-blocking for merge)

dq18:
Awesome feedback guys, I will get started to make the next version.
- I hid the footer for simplicity for a test version. I think that I can easily keep it but remove some chips (I think we only need the one with time).
- I'm not sure how to do zoom on each proof, but what I could do is to display it bigger when clicking on it (to be close to full screen) and have the dialog "Yes/No" somewhere at the bottom. WIll think a bit more about that.
- The time of the proof is kept. The dialog fetches the last 10 proofs with their info. I just reuse the date info for the price addition form. If you want a different behavior, let me know
- Having a separate proof upload page would be nice indeed, right now you have to go into any "add price" mode for that -> another PR :)
- "Proof uploaded" will be changed, along other hard-coded strings and other translations keys (will need some deep review though)

dq18:
- Added feature for AddPriceMultiple
- Added ProofFooter in a minimal mode (just the time chip)
image
- When clicking on a proof to choose, it now extends to the whole screen (in width because we keep ratio)
image

I'm removing the WIP status, who validates this can merge too :)

teolemon (approved):
I'm validitating because I want this badly ;-) @raphodn you can merge if that's ok technically. 
We'll iterate on this one, until it becomes truly addictive.

src/components/ExistingProofDialog.vue
```diff
@@ -0,0 +1,99 @@
...
+    },
+    close() {
+      this.$emit('close')
+    },
+  }
+}
+</script>
+
+<style scoped>
+.d-flex {
```
raphodn, line 96: why this style ? `d-flex` should work out of the box : https://vuetifyjs.com/en/styles/display/#display

src/components/ExistingProofDialog.vue
```diff
@@ -0,0 +1,99 @@
...
+      <v-divider></v-divider>
+      <v-card-text>
+        <v-row>
+          <v-col cols="6" sm="6" md="3" v-for="proof in userProofList" :key="proof">
+            <ProofCard :proof="proof" :reducedProofFooter="true" :isSelectable="true" :isSelected="selectedProof === proof" @proofSelected="selectProof"></ProofCard>
+          </v-col>
+        </v-row>
+    </v-card-text>
+    </v-card>
+    <v-bottom-sheet v-model="showConfirmationPopup">
```
raphodn, line 16: currently the confirmation popups (delete a price, delete a proof) are `v-dialog`. I would have kept this instead of creating a new behavior with the `v-bottom-sheet` + complexifying the `ProofCard` when selected :thinking:

src/components/ExistingProofDialog.vue
```diff
@@ -0,0 +1,99 @@
+<template>
+  <v-dialog>
+    <v-card>
+      <v-card-title>
+        {{ $t('ExistingProofDialog.ChooseLatestProofs') }} <v-btn style="float:right;" variant="text" density="compact" icon="mdi-close" @click="close"></v-btn>
+      </v-card-title>
+      <v-divider></v-divider>
+      <v-card-text>
+        <v-row>
+          <v-col cols="6" sm="6" md="3" v-for="proof in userProofList" :key="proof">
```
raphodn, line 10:
```suggestion
          <v-col cols="12" sm="6" md="3" v-for="proof in userProofList" :key="proof">
```
On small screens I think we should just show the list in 1 column. Also avoids having the zoomed image when selected, not really needed as the user can zoom already with 2 fingers

src/components/ProofFooter.vue
```diff
@@ -13,9 +13,9 @@
       </span>
       
     </v-chip>
-    <PriceCountChip :count="proof.price_count" :withLabel="true"></PriceCountChip>
+    <PriceCountChip v-if="!reducedProofFooter" :count="proof.price_count" :withLabel="true"></PriceCountChip>
```
raphodn, line 16: - why hide the price count ? seems a useful info to display
- and I'd rename `reducedProofFooter` to `hideProofDelete` (mecanism similar to other components)

src/views/AddPriceSingle.vue
```diff
@@ -169,12 +173,15 @@
...
                   :loading="createProofLoading">
                 </v-file-input>
-                <p v-if="proofFormFilled && !createProofLoading" class="text-green mt-2 mb-2">
+                <p v-if="proofFormFilled && !createProofLoading && !proofSelectedMessage" class="text-green mt-2 mb-2">
                   <i>{{ $t('AddPriceSingle.PriceDetails.ProofUploaded') }}</i>
                 </p>
                 <p v-if="!proofFormFilled && !createProofLoading" class="text-red mt-2 mb-2">
                   <i>{{ $t('AddPriceSingle.PriceDetails.UploadProof') }}</i>
                 </p>
+                <p v-if="proofFormFilled && proofSelectedMessage" class="text-green mt-2 mb-2">
```
raphodn, line 182: rather move this in the first `<p>` and add a v-if inside (avoids having 3 `<p>`)

src/components/ProofCard.vue
```diff
@@ -32,3 +50,20 @@ export default {
   }
 }
 </script>
+
+<style scoped>
+.highlighted {
```
raphodn, line 55: never used ?

raphodn (approved):

src/components/ExistingProofDialog.vue
dq18: That was to be able to see the whole proof before validating. With a `v-dialog`, one can't see the proof before validating

raphodn: v-dialog has a fullscreen option : https://vuetifyjs.com/en/api/v-dialog/#props-fullscreen

src/views/AddPriceSingle.vue
dq18: Since there is already a v-if in the first `<p>` element, I don't think that would simplify. What I could do is group the 3 `<p>` together with a computed method

src/components/ProofFooter.vue
dq18: the first `v-chip` for proof type is also hidden, so it would be named `hideProofTypeDelete`, but I suppose we should decide first what chip to keep before naming this boolean

src/components/ExistingProofDialog.vue
dq18: indeed

src/components/ExistingProofDialog.vue
raphodn: But honestly for a v1 I wouldn't even have a confirmation dialog (it's first and foremost for deletion, here it's... friction ?)
- worst case is that the user goes back to the "select a proof" 
- the user can zoom with 2 fingers
- 1 row of proofs on small screens

src/components/ProofFooter.vue
raphodn: why hide the proof type then ? :upside_down_face:

src/views/AddPriceSingle.vue
raphodn: currently
Screenshot from 2024-02-25 23-40-24

what I meant
Screenshot from 2024-02-25 23-41-32

which one is simpler ? ^^

src/components/ExistingProofDialog.vue
dq18: Alright

src/components/ProofCard.vue
dq18: Not with the current implementation of showing the card in fullscreen when selected. However if I remove that implementation, do we want to highlight or just show the confirmation dialog?

src/components/ProofFooter.vue
dq18: Taking a lot of space, and info not needed to in this case

src/components/ProofCard.vue
raphodn: neither ? proof simply selected on click

src/components/ProofFooter.vue
raphodn: I disagree, I wouldn't want to add GDPR proof here.
But in any case this is not important, you can keep it as is, it was just to simplify the logic

src/views/AddPriceSingle.vue
dq18: Alright, though I still think one <p> and a computed method is better :) But that can come in the future

src/components/ProofCard.vue
dq18: Alright. I commited the changes

raphodn:
merged ! made some last small changes : 
- make the dialog header "fixed"
- rename the dialog from "existing" to "recent" (to match with locations ; though it's very similar to "latest")

Issue #371, opened 2024-02-22
Deleted proof takes up space


ArturLange:
### What
After deleting a proof on user's dashboard, the proof still takes up space

### Steps to reproduce the behavior:
1. Go to https://prices.openfoodfacts.org/app/dashboard/proofs
2. Delete a proof
3. Proof is deleted but still takes up space

### Expected behavior
I would expect the layout to change and for the proofs to move to a vacant space.

### Screenshots
![obraz](https://github.com/openfoodfacts/open-prices-frontend/assets/10730549/e34e9192-693a-4273-a3ed-56ed1865fdb5)

### Platform (Desktop, Mobile, Hunger Games)
 - OS: Desktop

raphodn:
Fixed in #378

PR #378, opened 2024-02-24
fix: delete proof display

dq18:
### What
- Created a store dedicated to proofs storing for now the list of user proofs in frontend with add and remove methods
- Changed `ProofDeleteChip` to update this store with remove
- Changed  `UserDashboardProofList` to use the proofs list from the store

### Further improvements
- Move the api calls in the store ? We probably will use the same functions in several pages

### Fixes bug(s)
- Deleted proof takes up space #371 
The original way of removing a proofCard by manipulating the DOM was giving issues and not recommended in Vue.
An alternative is to "bubble" an emit event from `ProofDeleteChip` to `UserDashboardProofList` which makes things harder to follow


src/store/proofs.js
```diff
@@ -0,0 +1,18 @@
+import { defineStore } from 'pinia';
```
raphodn, line 1: Cool to start storing data in the store ! For instance #141

raphodn: But why in a seperate store ? Also, these are user proofs so could be under the user object, we could want to persist, and it avoid importing multiple stores, no ?

(sure in big apps they probably split stuff, but we don't need to introduce this complexity this soon, do we ?)

dq18: The idea of this PR was also to start a discussion on separating stores, what should really be separated and how much data should be stored (if we do prices for example, there might be too much data).
I don't know the best practice, one thought was to have some kind of matching with backend models, so separate stores for User, Proof, Prices, Location, Stats...

raphodn: we can start the discussion once our store feels too big or complex ^^
I think we're far from that situation

dq18: Cleaned up to use the unique store. Should be ready to merge

src/views/UserDashboardProofList.vue
```diff
@@ -67,8 +66,9 @@ export default {
       this.userProofPage += 1
       return api.getProofs({ owner: this.username, page: this.userProofPage })
         .then((data) => {
-          this.userProofList.push(...data.items)
-          this.userProofTotal = data.total
+          data.items.forEach(proof => this.appStore.addProof(proof))
+          this.appStore.user.proofs.sort((a, b) => new Date(b.created) - new Date(a.created))
+          this.userProofTotal = this.appStore.user.proofs.length
```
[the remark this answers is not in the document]
raphodn: a solution would be to keep the api response format in the store. or have the total in another key :thinking:

raphodn (approved):
Awesome !

PR #389, opened 2024-02-27
fix(product): fix add OFF button if product not found

### LINKED ISSUE #177: Product not found : add a link to OFF
### Story

When a new product is scanned but not found in our database, the product detail page shows a message, but no link to OFF

raphodn:
### What

In the product detail page, when the product is not found, we just show a message in red.

This PR does a few things : 
- adds a button to OFF
- changes the title to "<product code>" instead of "unknown product name"

Closes #177

### Screenshot

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/components/OpenFoodFactsButton.vue
+++ b/src/components/OpenFoodFactsButton.vue
@@ -20,7 +20,10 @@ export default {
   },
   methods: {
     getOFFUrl() {
-      return `${this.OFF_URL}/${this.type}/${this.value}`
+      if (this.type && this.value) {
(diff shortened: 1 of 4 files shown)


teolemon (commented):
Could we change the string to "Add this product to Open Food Facts" ?
Also, in some cases, this will be an OPF/OBF/OPFF product, so ideally several buttons

raphodn:
> Could we change the string to "Add this product to Open Food Facts" ?

ok !

> Also, in some cases, this will be an OPF/OBF/OPFF product, so ideally several buttons

I think it requires a seperate discussion, I don't want to add to many buttons right now

PR #395, opened 2024-02-28
refactor: use ProofCard everywhere

raphodn:
### What

We have duplicated code between ProofCard, PriceProof & ProofDeleteChip.

Added a new prop `hideProofHeader`, so that the ProofCard can be used everywhere !


dq18 (approved):
Awesome!

PR #396, opened 2024-02-28
feat(products): display popup with product labels

raphodn:
### What

Similar to #369

When the user clicks on the new "product labels count chip", a popup opens with the list of the labels.

### Screenshot

image


dq18 (approved):
looks good

PR #397, opened 2024-03-02
feat: allow using multiple currencies in Add Prices form

### LINKED ISSUE #359: Support setting the currency
As far as I can see, it is currently not possible to change the currency from EUR to anything else.

I quickly checked the backend and at least in the model, the currency is present: https://github.com/openfoodfacts/open-prices/blob/main/app/models.py#L130 although I'm not sure what the currencies list is containing exactly: https://github.com/openfoodfacts/open-prices/blob/main/app/enums.py#L5

I never contributed to an OFF project before, I don't know how you deal with such thing, does the backend have to tell the front-end which currencies are available, or do you want to go fast and hardcode them / load them from an external source in the front-end too?

In any case, it would be nice to see this supported.

Cheers,

dq18:
### What
- Following #406 
- Allow user to to choose among favorite currencies when adding a price
- Add price form will always have the last currency used as default

### Screenshot
- In AddSinglePrice and AddMultiplePrice pages, there is now a icon to change currency. Clicking on it will open a dialog to change among favorite currencies.
image
image

## What needs to be improved
- cleanup and rework UI
- Store form state when using routing to Settings (by extension, if user goes to any other route, refreshes...)

### Fixes issue
- closes #359


src/views/AddPriceMultiple.vue
```diff
@@ -246,9 +246,26 @@
               ></v-text-field>
             </v-col>
           </v-row>
-          <div class="d-inline">
+          <div class="d-flex">
             <v-checkbox v-model="productPriceForm.price_is_discounted" :label="$t('AddPriceSingle.PriceDetails.Discount')" hide-details="auto"></v-checkbox>
+            <v-checkbox v-model="isChangeCurrency" :label="$t('AddPriceSingle.PriceDetails.ChangeCurrency')" hide-details="auto"></v-checkbox>
```
raphodn, line 251: i feel it's misleading here to have a checkbox in a form, that is for a field not related to the form.

raphodn: alternative would be : 
- if the user has 1 currency set in its settings, then keep the current behavior, and add a sentence / info popup to explain how to change / add currency
- if the user has multiple curencies, then allow choosing with a dropdown

src/views/AddPriceMultiple.vue
raphodn, line 1: can we seperate into 2 PRs ? :pray: 

this PR has 2 distinct features here : 
- selecting multiple currencies in the settings
- choosing a currency in the form

src/store.js
```diff
@@ -11,6 +11,7 @@ export const useAppStore = defineStore('app', {
       recent_locations: [],
       language: localStorage.getItem('user-locale') || import.meta.env.VITE_DEFAULT_LOCALE,  // 'en'
       country: import.meta.env.VITE_DEFAULT_COUNTRY,  // 'FR',
+      currencies: [import.meta.env.VITE_DEFAULT_CURRENCY],  // ['EUR'],
```
raphodn, line 14:
```suggestion
      favorite_currencies: [import.meta.env.VITE_DEFAULT_CURRENCY],  // ['EUR'],
```
of "often used" or something like this ? both in the variable and in the UI

src/views/AddPriceSingle.vue
```diff
@@ -430,7 +447,11 @@ export default {
         this.setLocationData(this.recentLocations[0])
       }
       this.addPriceSingleForm.price_per = this.categoryPricePerList[0].key // init to 'KILOGRAM' because it's the most common use-case
-      this.addPriceSingleForm.currency = this.appStore.user.last_currency_used
+      this.addPriceSingleForm.currency = this.appStore.getUserLastCurrencyUsed
+      this.isChangeCurrency = false
+    },
+    goToSettings() {
```
raphodn, line 453: I'm not sure this is a good idea to put a link to the settings here. the form will be lost, the user doesn't know that. Alternatives :
- clear message that form will be lost
- allow changing the currency settings in a popup (without going to the settings)

dq18:
@raphodn I updated with a better UI. Let me know what you think.
This PR has commits from #406 but I will rebase once the first PR is merged.

src/components/ChangeCurrencyDialog.vue
```diff
@@ -0,0 +1,65 @@
...
+        <v-col cols="12">
+          <v-btn class="mb-2" size="small" prepend-icon="mdi-cog-outline" @click="goToSettings">
+            <span>{{ $t('ChangeCurrencyDialog.AddCurrencies') }}</span>
+          </v-btn>
+          <p class="text-caption text-warning">
+            <i>{{ $t('ChangeCurrencyDialog.AddCurrenciesWarning') }}</i>
+          </p>
+        </v-col>
+        <v-col cols="12">
+          <v-btn class="mt-2" color="success" @click="selectedCurrency">{{ $t('UserSettings.Save') }}</v-btn>
```
raphodn, line 24: - not exactly "Save", more "Select" ? even better "Select EUR". 
- nothing created here, i wouldn't use the "success" color (like barcode scanner)

src/views/AddPriceMultiple.vue
```diff
@@ -232,7 +232,12 @@
                 min="0"
                 hide-details="auto"
                 :suffix="productPriceForm.currency"
-              ></v-text-field>
+                >
+                  <template v-slot:prepend>
+                    <!-- image from https://www.svgrepo.com/svg/32717/currency-exchange -->
+                    <img src="/currency-exchange-svgrepo-com.svg" class="icon-info-currency" @click="changeCurrencyDialog = true" />
```
raphodn, line 238: i guess there's no similar mdi icon :(

raphodn: i would add a `title` (can be done in a seperate PR and add accessibility elsewhere)

src/i18n/locales/en.json
```diff
@@ -120,6 +121,13 @@
 		"Yes": "Yes",
 		"No": "No"
 	},
+	"ChangeCurrencyDialog": {
+		"Title": "Choose one of favorite currencies",
```
raphodn, line 125:
```suggestion
		"Title": "Choose one of your favorite currencies",
```
?

raphodn (approved):
LGTM !

Issue #363, opened 2024-02-21
Allow users to mark a receipt as private


raphael0202:
Some receipts contain personal information (fidelity card, user first and last name,...).
The API supports indicating a proof as private by setting `is_public=false`. A checkbox should be added on the frontend to allow to mark the receipt proof as private.

PR #401, opened 2024-03-03
feat: add update PATCH proof in api

dq18:
### What
Thanks to https://github.com/openfoodfacts/open-prices/issues/236, it is now possible to update a proof fields (`type` and `is_public`)
Created the endpoint on the frontend side

Part of  #363


raphodn (approved):
LGTM !

Issue #402, opened 2024-03-03
Allow users to edit/update their prices


raphodn:
### Story

Frontend version of https://github.com/openfoodfacts/open-prices/issues/181

- [x] simple edit form
- [x] delete button

PR #403, opened 2024-03-03
feat(api): new updatePrice method

raphodn:
### What

Thanks to https://github.com/openfoodfacts/open-prices/issues/181 there is now a backend endpoint to update (PATCH) a price.

Add a new `updatePrice()` method in the frontend api

Part of #402


dq18 (approved):
Nice!

PR #404, opened 2024-03-03
feat: clickable proof chip to set RECEIPT as private

dq18:
### What
Thanks to #401 there is an API endpoint to update some proof fields. User can now see in Proof page a chip for RECEIPT type proofs.

image

This chip is clickable and will toggle between "Public" and "Private"

Part of #363

### DIFF
--- a/src/components/ProofFooter.vue
+++ b/src/components/ProofFooter.vue
@@ -15,7 +15,9 @@
     </v-chip>
     <PriceCountChip :count="proof.price_count" :withLabel="true" @click="goToProof()"></PriceCountChip>
     <RelativeDateTimeChip :dateTime="proof.created"></RelativeDateTimeChip>
+    <ProofIsPublicChip v-if="proof.type === 'RECEIPT'" :proof="proof"></ProofIsPublicChip>
     <ProofDeleteChip v-if="!hideProofDelete && userCanDeleteProof" :proof="proof"></ProofDeleteChip>
+
   </div>
 </template>
 
@@ -29,7 +31,9 @@ export default {
   components: {
     'PriceCountChip': defineAsyncComponent(() => import('../components/PriceCountChip.vue')),
     'RelativeDateTimeChip': defineAsyncComponent(() => import('../components/RelativeDateTimeChip.vue')),
-    'ProofDeleteChip': defineAsyncComponent(() => import('../components/ProofDeleteChip.vue'))
+    'ProofDeleteChip': defineAsyncComponent(() => import('../components/ProofDeleteChip.vue')),
+    'ProofIsPublicChip': defineAsyncComponent(() => import('../components/ProofisPublicChip.vue'))
+
   },
   props: {
     'proof': null,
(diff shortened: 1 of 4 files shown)


src/components/ProofFooter.vue
```diff
@@ -15,7 +15,9 @@
     </v-chip>
     <PriceCountChip :count="proof.price_count" :withLabel="true" @click="goToProof()"></PriceCountChip>
     <RelativeDateTimeChip :dateTime="proof.created"></RelativeDateTimeChip>
+    <ProofIsPublicChip v-if="proof.type === 'RECEIPT'" :proof="proof"></ProofIsPublicChip>
```
raphodn, line 18: my 2 cents : in terms of UX, I wouldn't mix tags and actions.
- currently there's info tags on the left, and a delete button on the right
- i would rather add an extra "edit" button : that would allow editing both the type & the public, inside a dialog
- and maybe put the delete inside it as well in the long term ? (or not, haven't given much thought on it yet)

raphodn: also, it's missing some kind of `userCanEditProof`

raphodn: also it's missing some kind of `userCanDeleteProof` or `isProofOwner`

raphodn: to go even further in the reflexion, this could be 2 PR : 
- 1 to display the public or private information (as an owner & as an other user)
- 1 to edit this information (if you are the owner)

dq18: I agree, we should rethink the proof page and the rest as the app becomes more complex. We are close to needing a big UI/UX research phase.

I am not sure I understand what you mean by:
> also it's missing some kind of userCanDeleteProof or isProofOwner
Do you mean we can use those methods outside the proof page?
Right now, only the user can see a proof that he uploaded and that has no price associated (thus can be deleted). A deletable proof is never seen by other users.
When checking a proof associated by a price by clicking the proof chip, we can just return a tooltip saying that it is private if not the owner. This just needs an assertion of `proof.owner` on click.

Following this PR, 2 more will follow:
- Set proof as private when adding from a "Add Price Receipt" form. Still hesitating if I should pass it in proofCreate or proofUpdate. Since proof gets uploaded right after selection, the parameter is_public has to be chosen by the user before the selection, which is counter-intuitive. My thought is to add a toggle or checkbox showing up once proof is uploaded to mark it as private.
- Disable non owners to see a private proof as mentioned above

src/components/ProofFooter.vue
```diff
@@ -29,7 +31,9 @@ export default {
   components: {
     'PriceCountChip': defineAsyncComponent(() => import('../components/PriceCountChip.vue')),
     'RelativeDateTimeChip': defineAsyncComponent(() => import('../components/RelativeDateTimeChip.vue')),
-    'ProofDeleteChip': defineAsyncComponent(() => import('../components/ProofDeleteChip.vue'))
+    'ProofDeleteChip': defineAsyncComponent(() => import('../components/ProofDeleteChip.vue')),
+    'ProofIsPublicChip': defineAsyncComponent(() => import('../components/ProofisPublicChip.vue'))
```
raphodn, line 35:
```suggestion
    'ProofIsPublicChip': defineAsyncComponent(() => import('../components/ProofIsPublicChip.vue'))
```

src/components/ProofFooter.vue
raphodn: as mentionned above, there should be a check `userCanEditProof`
for example, in the latest prices, any user can see the price proofs

dq18: Alright, it's my point 2 

> Disable non owners to see a private proof as mentioned above

I will start working on that

PR #405, opened 2024-03-05
feat: Add price proof private in Add Price

dq18:
### What
- Allow user to set Private for a Reicept proof directly in Add Price.
image

- It calls the update proof PATCH endpoint. Since for now a proof is uploaded directly when choosing an image, to pass is_public on creation requires user to choose this before looking at the image, not very intuitive


src/views/AddPriceMultiple.vue
```diff
@@ -50,18 +50,37 @@
...
               <p v-if="!proofFormFilled && !createProofLoading" class="text-red mt-2 mb-2">
                 <i>{{ $t('AddPriceSingle.PriceDetails.UploadProof') }}</i>
               </p>
               <p v-if="proofType === 'RECEIPT'" class="text-caption text-warning">
                 <i>{{ $t('AddPriceMultiple.ProofDetails.ReceiptWarning') }}</i>
+                <i>{{ $t('AddPriceMultiple.ProofDetails.PrivateWarning') }}</i>
               </p>
             </v-col>
-            <v-col v-if="proofFormFilled">
+            <v-col  cols="12" sm="12" md="4" v-if="proofFormFilled">
```
raphodn, line 64:
```suggestion
            <v-col cols="12" md="4" v-if="proofFormFilled">
```

raphodn: also, when switching to md, the whole section switches as well, so there is not that much space for the proof image + toggle
image

src/views/AddPriceMultiple.vue
```diff
@@ -50,18 +50,37 @@
...
               <v-img :src="proofImagePreview" style="max-height:200px"></v-img>
+              <div class="proofIsPublic-switch-container" v-if="proofFormFilled && (proofType === 'RECEIPT')">
```
raphodn, line 66: I would have moved this to a row below.
- warning text
- then switch (with label next to it)

raphodn: example : (juste avec des row et des col, plus besoin de css custom)

image

PR #408, opened 2024-03-06
feat: new ordering choice by price (lowest first)

raphodn:
### What

In the pages with a list of prices, add a new ordering by "cheapest first"

### Screenshot

image

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -20,8 +20,9 @@ export default {
     { key: '-price_count', value: 'Number of prices', icon: 'mdi-tag-multiple-outline' },
   ],
   PRICE_ORDER_BY_LIST: [
-    { key: '-created', value: 'Addition date', icon: 'mdi-clock-outline' },
+    { key: 'price', value: 'Price (cheapest first)', icon: 'mdi-currency-eur' },
     { key: '-date', value: 'Price Date', icon: 'mdi-calendar' },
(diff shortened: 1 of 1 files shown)


src/constants.js
```diff
@@ -20,8 +20,9 @@ export default {
     { key: '-price_count', value: 'Number of prices', icon: 'mdi-tag-multiple-outline' },
   ],
   PRICE_ORDER_BY_LIST: [
-    { key: '-created', value: 'Addition date', icon: 'mdi-clock-outline' },
+    { key: 'price', value: 'Price (cheapest first)', icon: 'mdi-currency-eur' },
```
dq18, line 23: I would put "Price: ascending order" as I have seen on several marketplace.
Also, would put another icon, as euro is not relevant outside EU. Maybe `order-numeric-ascending`, since I'm sure the descending option will come soon :)

dq18 (approved):

src/constants.js
raphodn: good idea for the icon :+1:

raphodn: I'll change to just "Price", and we can open an issue to figure out the best wording :)

dq18: Hmm true, my bad. I rarely go to marketplaces and when I do they are in German with autotranslation to English. In German it is "Aufsteigend" and "Absteigend" which translates to "Ascending" and "Descending"

PR #409, opened 2024-03-06
refactor(l10n): allow translation of price ordering options

raphodn:
### What

Currently the price ordering values are hardcoded in `constants.js`

Refactored to allow translations


dq18 (approved):
Awesome!

PR #411, opened 2024-03-06
refactor(l10n): allow translation of filtering options

raphodn:
### What

Similar to #410 & #409 but for the filter(s)


dq18 (approved):
Awesome too :)

PR #419, opened 2024-03-08
fix(i10n): language dropdown and locale load at mount

dq18:
### What
Due to recent language cleanup PR #337 some issues were raised
- UserSettings language dropdown was not showing the current locale name but code
image
- Refresh or direct URL would default text to English

### Changes
- Small fix in UserSettings
- Added call of `changeLanguage` before App mount


raphodn:
i18n = internationalization
l10n (and not i10n) = localization
:wink:

Issue #438, opened 2024-03-13
Proof info is not updated in Dashboard


dq18:
### Steps to reproduce the behavior:
1. Go to Dashboard My proofs page and check info for a proof
2. Go to add price and use the proof to create new prices
3. Go back to My proofs to see that there is no change

### Expected behavior
Info should be updated

### Solution
- proof list in store will only be updated with new proofs but not check that their properties have changed

PR #442, opened 2024-03-14
feat(price): new action menu if user is owner (to delete, and soon edit)

raphodn:
### What

New `PriceActionMenuButton` component
Move Delete action to `PriceDeleteConfirmationDialog`

### Screenshot

Screenshot from 2024-03-14 16-03-10


src/components/PriceDeleteConfirmationDialog.vue
```diff
@@ -40,9 +42,26 @@ export default {
...
+        .then((response) => {
+          // if response.status == 204
+          this.loading = false
+          this.removePriceCard()
+          this.$emit('delete')
+          this.closeDialog()
+        })
+    },
+    removePriceCard() {
+      document.getElementById(`price_${this.price.id}`).remove()
```
dq18, line 64: Same issue we had with proofs
image
An alternative is to keep in store but find a way to update the store when the database is updated to avoid something like #438

dq18 (approved):

src/components/PriceDeleteConfirmationDialog.vue
raphodn: indeed, created an issue #449

PR #455, opened 2024-03-16
fix: remove errors on Proof card emits

raphodn:
### What

Errors in the console regarding `ProofCard`
- fixed by adding the `emits` param to the component
- improved the wording

### Screenshot

image


src/components/UserRecentProofsDialog.vue
```diff
@@ -35,6 +37,7 @@ export default {
       selectedProof: null,
     }
   },
+  emits: ['recentProofSelected', 'close'],
```
raphodn, line 40: something to add in other components as well (in another refactoring PR)

Issue #551, opened 2024-04-21
Allow comma as decimal separator


tuukka:
### What
When entering a price, my phone keyboard offered a comma as usual for the decimal separator.

### Steps to reproduce the behavior:
1. Start to enter a price.
2. Type in a price that uses comma as the decimal separator.
3. There was no indication of an error but the price became invisible and the Upload button stayed disabled.

### Expected behavior
Accept comma as the decimal separator, or at least give an indication of what's wrong.

### Platform
 - Android
 - Firefox

PR #554, opened 2024-04-24
fix: accept comma as decimal separator in prices (#551)

tuukka:


src/components/PriceInputRow.vue
```diff
@@ -59,6 +61,18 @@ export default {
...
       this.priceForm.currency = currency
     },
+    fixComma(input) {
+      return input.replace(/,/g, '.');
+    },
+  },
+  computed: {
+    priceRules() {
+      return [
+        value => !!value && !!value.trim() || 'Price is required',
```
raphodn, line 71: needs translations :)

src/components/PriceInputRow.vue
```diff
@@ -2,11 +2,12 @@
   <v-row>
     <v-col :cols="priceForm.price_is_discounted ? '6' : '12'" sm="6">
       <v-text-field
-        v-model="priceForm.price"
+        :model-value="priceForm.price"
+        @update:model-value="newValue => priceForm.price = fixComma(newValue)"
```
raphodn, line 6: :+1:

src/components/PriceInputRow.vue
```diff
@@ -2,11 +2,12 @@
...
         :label="priceForm.price_is_discounted ? $t('PriceForm.LabelDiscounted') : $t('PriceForm.Label')"
-        type="number"
+        type="text"
```
raphodn, line 8: the smartphone numeric keyboard is still displayed ?

raphodn: maybe add a pattern ? I tested and I could input letters, sure there is the error message but better to restrict in the first place no ?

raphodn: `pattern="[0-9]+([\.,][0-9]+)?"`

src/components/PriceInputRow.vue
```diff
@@ -59,6 +61,18 @@ export default {
...
+    fixComma(input) {
+      return input.replace(/,/g, '.');
+    },
+  },
+  computed: {
+    priceRules() {
+      return [
+        value => !!value && !!value.trim() || 'Price is required',
+        value => !isNaN(value) || 'Input a number',
+        value => Number(value) >= 0 || 'Price may not be negative',
```
raphodn, line 73: add a rule with a maximum of 2 decimals ?

raphodn: fyi i added a PR with 2 extra tests in the backend : https://github.com/openfoodfacts/open-prices/pull/279

raphodn (commented):
Thanks for the PR ! I added some comments, only tested locally on my laptop, not on a smartphone. Almost there :)

src/components/PriceInputRow.vue
tuukka: I'm not sure I did it right, I didn't see style instructions.

src/components/PriceInputRow.vue
tuukka: Yes, the numeric keyboard is still displayed because of `inputmode="decimal"`.

The `pattern` didn't work in a useful way, at least on Firefox. Also, I think it's better to show an error rather than silently drop characters from the user input, e.g. "-5.99 bananas (return)" is good to catch rather than silently convert to "5.99". Same for "26 130" which may be two numbers etc.

I added an error message regarding spaces.

src/components/PriceInputRow.vue
tuukka: I added the rule of (max) 2 decimals, as it was simple and it's important to catch if someone uses a thousands separator "1.000".

tuukka:
Thank you for the valuable comments. This may seem to be a lot of code for a simple number field but on the other hand, it's perhaps the most important UI functionality of Open Prices :relaxed:

raphodn (approved):
Thanks for the swift changed :100:

PR #557, opened 2024-04-25
feat(proof): add the price CTA button in the proof detail page

raphodn:
### What

Following #375 & #521

Add the "Add price" CTA button on the proof detail page

### Stuff missing

- doesn't fetch the proof image
- doesn't check if the user is the proof owner (but the CTA button is only displayed on pages where the user is owner of the proof, and it will have an error in the backend anyway)

### Screenshot

image


odin-h:
Does it need to check if the user is the owner of the proof image? What if someone who's not the owner sees a price tag in the image that the original uploader forgot to add?

raphodn:
> Does it need to check if the user is the owner of the proof image? What if someone who's not the owner sees a price tag in the image that the original uploader forgot to add?

Currently, the backend server expects that both the price owner and the proof owner should be the same (except if they are moderators, but we don't use this feature yet in the frontend).

odin-h:
Oh, okay 👍

PR #584, opened 2024-05-23
feat(proof): fetch and insert proof data to 'add single price' page when using 'add the price' button

odin-h:
### What
- Solves the missing proof image from https://github.com/openfoodfacts/open-prices-frontend/pull/557 and also inserts the proof created date to the form.
 _It uses the date the proof was uploaded, so might not always be the correct date for when the picture was taken_

### Fixes bug(s)
- 

### Part of 
-

### DIFF
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -412,17 +412,27 @@ export default {
     showUserRecentProofs() {
       this.userRecentProofsDialog = true
     },
-    handleProofSelected(proofId) {
-      this.addPriceSingleForm.proof_id = proofId
+    handleProofSelected(proof) {
+      this.addPriceSingleForm.proof_id = proof.id
+      this.addPriceSingleForm.date = new Date(proof.created).toISOString().split('T')[0]
+      this.proofImagePreview = this.getProofUrl(proof)
       this.proofSelectedSuccessMessage = true
       this.proofSelectedMessage = true
     },
     handleRecentProofSelected(selectedProof) {
-      this.handleProofSelected(selectedProof.id)
+      this.handleProofSelected(selectedProof)
       this.proofImagePreview = this.getProofUrl(selectedProof)
     },
     getProofUrl(proof) {
       return `${import.meta.env.VITE_OPEN_PRICES_APP_URL}/img/${proof.file_path}`
+    },
+    getProofById(proofId) {
+      this.loading = true;
+      api.getProofById(proofId)
+        .then(proof => {
+          this.handleProofSelected(proof);
+          this.loading = false;
+        });
     },
     newProof(source) {
       if (source === 'gallery') {
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -383,7 +383,7 @@ export default {
         this.setProductCode(this.$route.query.code)
       }
     } else if (this.$route.query.proof) {
-      this.handleProofSelected(this.$route.query.proof)
+      this.getProofById(this.$route.query.proof);
     }
     this.initPriceSingleForm()
   },
--- a/src/views/AddPriceSingle.vue
+++ b/src/views/AddPriceSingle.vue
@@ -382,8 +382,8 @@ export default {
       else {
         this.setProductCode(this.$route.query.code)
       }
-    } else if (this.$route.query.proof) {
-      this.getProofById(this.$route.query.proof);
+    } else if (this.$route.query.proof_id) {
+      this.getProofById(this.$route.query.proof_id)
     }
     this.initPriceSingleForm()
   },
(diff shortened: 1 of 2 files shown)


src/views/AddPriceSingle.vue
```diff
@@ -412,18 +412,28 @@ export default {
     showUserRecentProofs() {
       this.userRecentProofsDialog = true
     },
-    handleProofSelected(proofId) {
-      this.addPriceSingleForm.proof_id = proofId
+    handleProofSelected(proof) {
+      this.addPriceSingleForm.proof_id = proof.id
+      this.addPriceSingleForm.date = new Date(proof.created).toISOString().split('T')[0]
```
raphodn, line 417: I'd rather keep it empty and let the user fill it in, no ? Or some kind of warning ?

src/views/AddPriceSingle.vue
```diff
@@ -412,18 +412,28 @@ export default {
     showUserRecentProofs() {
       this.userRecentProofsDialog = true
     },
-    handleProofSelected(proofId) {
-      this.addPriceSingleForm.proof_id = proofId
+    handleProofSelected(proof) {
```
raphodn, line 415: the problem here is that above in `mounted` we only have the proof_id, so this will not work...

raphodn: can be fixed by keeping `proofId`, and removing the date init.

raphodn: and add an extra param to fetch or not the proof image ? because if coming from the `UserRecentProofsDialog` then it we be called even though we already have the image (line 424)

src/views/AddPriceSingle.vue
odin-h: When setting a recent proof or proof from URL parameter ID I definitely intended the date to be automatically set like it already does when uploading a new proof, but maybe not to the proof upload date like it does here. Preferably it should set the date to when the picture was taken instead of today's date like the current running version of the site does when selecting a proof right now. However, without adding a column for file creation time in the database, it probably have to use some potentially expensive sql queries to determine the date based on other prices uploaded to the proof already, if it even has any price entries at all. I've tried to download proof images I uploaded with exif data that has been compressed and process into mebp files, but I was not able to find any of the original exif data or dates. Only the modification date of when I uploaded the file to Open Prices. It could be that the file's exif date is stripped when serving the file through the URL (and still available internally), but it appears that maybe that data is lost in the compression process.

src/views/AddPriceSingle.vue
odin-h: Oh, I forgot to change the function used in `mounted` to `getProofById` when copying the code over to GitHub. Did that now.

odin-h: Not sure what you mean by it being called twice when using the `UserRecentProofsDialog` though. It's only calling it once for me. Maybe I'm misunderstanding you or something is wrong with my testing environment. I couldn't get image uploading to work, uploaded pics just give me 404 error when I try to access the upload path, so I had to do a few tweaks to work around that by using remote URLs instead.

src/views/AddPriceSingle.vue
odin-h: The date in the json response makes even less sense. It's before I even took the image. I'm assuming it's using a different time zone and it's actually 17:00 reflecting the date of the downloaded image:
```json
{
  "id": 6644,
  "file_path": "0007/2rO9xlT7Nn.webp",
  "mimetype": "image/webp",
  "type": "PRICE_TAG",
  "price_count": 1,
  "owner": "odinh",
  "created": "2024-05-21T15:00:46.692372Z"
}
```
Here's a screenshot showing the metadata of both the downloaded image and the original:
image

raphodn: yeah I think the server is on a strange timezone or something.
maybe we could add a field "date_taken" ? though sometimes we might not have this info, if an image is uploaded without the related metadata

odin-h: Maybe changing `created` in the json response from a offset-naive to a offset-aware timestamp would be a good idea to do instead, so it includes the timezone offset? And regarding the file metadata, could maybe make it default to the file's creation date instead, if `date_taken` is not available? Most files usually have a creation date even if they don't have the `date_taken` EXIF value. Another alternative would be to extract it from the filename like you could for the file in my example:
```
20240521_162617.jpg
```

src/views/AddPriceSingle.vue
raphodn: my bad there doesn't seem to be duplicate loading :ok_hand:

raphodn:
ok i pushed a quick commit for linting/cleanup, will merge :100:

src/views/AddPriceSingle.vue
raphodn: linked issues in the backend : 
- https://github.com/openfoodfacts/open-prices/issues/314
- https://github.com/openfoodfacts/open-prices/issues/107

PR #589, opened 2024-05-26
feat(proof): recent proof dialog: add load more button, filter list depending on receipt or price tag

odin-h:
### What
- Adds a load more button at the bottom of the recent proofs list in the dialog box used when adding prices.
Also, filters the proof type when fetching proofs based on if you're adding receipts or regular price tags.

### Fixes bug(s)
- 

### Part of 
-

### DIFF
--- a/src/components/UserRecentProofsDialog.vue
+++ b/src/components/UserRecentProofsDialog.vue
@@ -13,6 +13,11 @@
             <ProofCard :proof="proof" :hideProofHeader="true" :hideProofActions="true" :readonly="true" :isSelectable="true" @proofSelected="selectProof" />
           </v-col>
         </v-row>
+        <v-row v-if="userProofList.length < userProofTotal" class="mb-2">
+          <v-col align="center">
+            <v-btn size="small" :loading="loading" @click="getUserProofs">{{ $t('ProofDetail.LoadMore') }}</v-btn>
+          </v-col>
+        </v-row>
       </v-card-text>
     </v-card>
   </v-dialog>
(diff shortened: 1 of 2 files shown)


src/components/UserRecentProofsDialog.vue
```diff
@@ -33,7 +38,8 @@ export default {
     return {
       userProofList: [],
       userProofTotal: null,
-      userProofPage: 1,
+      userProofPage: 0,
+      userProofType: this.$route.path.endsWith('/receipt') ? 'RECEIPT' : 'PRICE_TAG',
```
raphodn, line 42: the issue here is that the component's behavior will rely on the url. it would be better to be an optional parameter (props)

raphodn (commented):
Awesome !

src/components/UserRecentProofsDialog.vue
raphodn: I fixed it here :) https://github.com/openfoodfacts/open-prices-frontend/pull/589/commits/97f9852cb537417c95f65b6d1b5e8b70771381cc

raphodn (approved):

src/components/UserRecentProofsDialog.vue
odin-h: Nice, thanks! 👍 

However, if I'm understanding it correctly now that this filter is in place, the 'add a price' button on a specific proof id page with the `RECEIPT` type won't show the selected proof in the recent proofs dialog box on the add price page anymore. That could maybe be resolved by combining all the 'add price' pages into the add multiple page instead and add a proof type selector on the page to choose between price tag and receipt?

PR #617, opened 2024-06-07
feat(location search): if the user searches by ID, then do a lookup with Nominatim

raphodn:
### What

Offer the possibility to search by node/way/relation ID

### Screenshot

image


raphodn:
fyi @monsieurtanuki i added a search by OSM id

PR #623, opened 2024-06-09
refactor(price add): change the ui of the discount toggle (switch instead of checkbox)

raphodn:
### What

Replace the discount checkbox with a switch

### Screenshot

|Before|After|
|---|---|
|image|image|


teolemon:
nice touch @raphodn

PR #633, opened 2024-06-18
Issue 530

nobeeakon:
### What
- UI changes to the currency selector for Add single price. Button to edit + selector, instead of open a dialog and be redirected to the settings page.

### Screenshot
image

### DIFF
--- a/src/components/PriceInputRow.vue
+++ b/src/components/PriceInputRow.vue
@@ -10,12 +10,7 @@
         hide-details="auto"
         :suffix="priceForm.currency"
         @update:model-value="newValue => priceForm.price = fixComma(newValue)"
-      >
-        <template #prepend-inner>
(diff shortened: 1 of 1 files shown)


raphodn:
Some recent changes to the UI : 
- #645
- a help tooltip next to the currency label : #649 

Not sure if we should add complexity in the proof/price add form. But rather redirect users to their settings page (currencies don't change very often...). And maybe one day have image detection, and or automatic currency setup depending on location

nobeeakon:
think it is easier for the user to just select the currency they want, rather than being redirected to another page in the site to update this info and having to refresh the page (were they may have already added some information) to get the latest values. Besides that, don't think it is adding a lot of complexity, as is mostly some UI elements, and a couple of lines of code to the logic.

PR #762, opened 2024-08-21
ci: Update labeler.yml to V5 syntax

teolemon:
### What
- ci: Update labeler.yml to V5 syntax

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,42 +1,54 @@
 # For reference, please look at https://github.com/actions/labeler
 
 GitHub Actions:
-- .github/**/*
+- changed-files:
+  - any-glob-to-any-file: '.github/**/*'
 
 Price addition:
-- 'src/views/AddPriceHome.vue'
-- 'src/views/AddPriceSingle.vue'
-- 'src/views/AddPriceMultiple.vue'
+- changed-files:
+  - any-glob-to-any-file: 'src/views/AddPriceHome.vue'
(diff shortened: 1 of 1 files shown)


raphodn:
closing (reverted #735 in https://github.com/openfoodfacts/open-prices-frontend/pull/763)

raphodn:
reopening after re-reverting #765

.github/labeler.yml
```diff
@@ -1,42 +1,54 @@
...
-- .github/**/*
+- changed-files:
+  - any-glob-to-any-file: '.github/**/*'
 
 Price addition:
-- 'src/views/AddPriceHome.vue'
-- 'src/views/AddPriceSingle.vue'
-- 'src/views/AddPriceMultiple.vue'
+- changed-files:
+  - any-glob-to-any-file: 'src/views/AddPriceHome.vue'
```
raphodn, line 9: shouldn't it be 1 line with an array of values ?

https://github.com/actions/labeler#basic-examples

teolemon: it works with my syntax on other repos 🤷

raphodn:
@teolemon doesn't seem to work ? see one of the latest PRs (for instance #769)

teolemon:
- instead of reverting, let's fix master so that the labeler works again. The "actions" tab is there for us
- also for release please, you're using slightly different conventions from the rest of the repos, so I need to have an extra look (but I think we should just fix the config in crowdin or release please, and missing a PR or 2 is not dramatic)

PR #783, opened 2024-08-22
chore: update categories

raphael0202:
- add new categories (spices, flours, aromatic plants, legumes)
- deduplicate categories before saving files
- format `filter_categories.py` using black
- use `get_taxonomy` with `force_download=True` and `download_newer=True` to make sure you have latest version
- add categories that were deleted from category list due to category taxonomy updates


raphodn (approved):
cool ! and i didn't know there was a difference in english between `vegetables` & `legumes` :sweat_smile:

PR #828, opened 2024-09-18
refactor(api): bump default page_size to 100

raphodn:
### What

The current frontend default is to load data by batches of 10.
The backend server allows 100.

For stats purposes, we bump the default page size to 100 (load more prices => show more locations on the map & points on the price chart)


raphodn:
100 was slowing things down.
reverted to 20 (https://github.com/openfoodfacts/open-prices-frontend/commit/f4fdf8b3001c9e7b6c967de7fcbabd0c0f7a50da), and finally to 25 (https://github.com/openfoodfacts/open-prices-frontend/commit/1d5fac1281bb181759101a5c4dee7ed3a879d406) to match a bit with OFF main website (50)

Issue #843, opened 2024-09-25
Revamp the landing page


raphodn:
### Story

Linked with #364 (make the root of the app https://prices.openfoodfacts.org/ instead of https://prices.openfoodfacts.org/app/ )

### Currently

(logged-in user)

![Screenshot from 2024-09-28 15-29-37](https://github.com/user-attachments/assets/489aa066-16bc-48f1-914e-bf6239a94ef6)

### Todo

- [x] move actions (the 3 existing card-buttons) to the header & side drawer
- [x] show some global stats at the top
- [ ] show some personal stats below
- [x] show latest prices (25, but reduce to 5 on small screens)
- [ ] links to about, ...
- [ ] some info about OFF ?
- [x] reduce the footer height : https://github.com/openfoodfacts/open-prices-frontend/commit/6d0536748591f1264e8872b10831d73e46589c0c

get inspiration from https://hunger.openfoodfacts.org for instance !

raphodn:
@serpico any feedback on the upcoming/new home page ?
https://prices.openfoodfacts.net/app/

serpico:
Maybe it's just an impression but on some pages the "cards" look too big ( but maybe there's a reason, like more space for upcoming data displayed on them ), like in https://prices.openfoodfacts.org/app/prices, much fewer cards for the same screen size compare to previously but it's not on the .net so it must be from before...however I have all the extra chip (3) enabled ( but I had that before, since they're available ) and I have small screen laptop 12.5" so it's  only 6 and half cards before it felt like 9 or more. But it's probably just a matter of getting used to it.

raphodn:
I share your feeling that the price cards size can be reduced, and the spacing between them as well 👌

raphodn:
v2 deployed (there still room for improvement :)

![image](https://github.com/user-attachments/assets/3241d1d4-efa5-41c0-bf09-3eaace9496a3)

PR #865, opened 2024-09-28
feat(Home): reorganize home page sections. move buttons to header

raphodn:
### What

See issue #843 

Improved Home Page
- show basic stats
- show 5 latest prices + button to see more
- Search & Add actions moved to the header (and they remain available in the top-left menu)

### Todo (here or next PR)

- more stats (locations, users)
- center "Latest prices" button

### Screenshot

||Before|After|
|---|---|---|
|Desktop|image|image|
|Mobile|image|image|

### DIFF
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -1,59 +1,29 @@
 <template>
   <h1 class="text-h5 mb-1">
-    <i18n-t keypath="Home.Welcome.Title" tag="span">
-      <template #name>
-        {{ APP_NAME }}
-      </template>
-    </i18n-t>
+    {{ $t('Home.Welcome.Generic') }}
   </h1>
-  <p>{{ $t('Home.Welcome.Subtitle') }}</p>
-  
+
   <br>
 
   <v-row>
-    <v-col cols="12" sm="8" md="6" lg="4">
-      <v-card
-        :title="$t('Home.SearchProduct')"
-        prepend-icon="mdi-magnify"
-        height="100%"
-        to="/search"
-      />
+    <v-col sm="6" md="4" lg="3">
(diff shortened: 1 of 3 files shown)


src/views/Home.vue
```diff
@@ -108,9 +81,19 @@ export default {
     if (this.$route.query.settingsSuccess === 'true') {
       this.settingsSuccessMessage = true
     }
+    this.getPrices()
     this.getTodayPriceCount()
   },
   methods: {
+    getPrices() {
+      this.loading = true
+      return api.getPrices({ size: 5 })
```
raphodn, line 90: How many prices should we display ? 
Maybe load 5, but display only 2 or 3 on small screens ?

teolemon: I'd load up to 25, adding "3 minutes ago", up to"24 hours ago" instead of Today's date

raphodn: ok will re-add the "relative time info" in the PriceCard in a different PR

raphodn: and I need to think about loading more data for bigger screens (different PR as well)

raphodn: bumped to 25 (and kept it to 5 on small screens) in this commit : https://github.com/openfoodfacts/open-prices-frontend/commit/588c97a9f6ed14fe73bd39b0f0d5351973758071

raphodn: > instead of Today's date

its the price's date that is displayed. sometimes prices are added for a different date.
indeed it can be confusing..

teolemon: It was more to make it more alive and exciting, I wasn't particularly confused by anything

Issue #364, opened 2024-02-21
Kill the permanent landing page, and turn it into a one time thing


teolemon:
### Problem

- Kill the permanent landing page, and turn it into a one time thing
- prices.openfoodfacts.org will be the app
- prices.openfoodfacts.org/about will be the existing landing page
- We have the promo in the app to read the landing page.

### Todo

- [x] new /about page with the existing content: #874
- [x] translate the page via Crowdin using the existing /fr page: #884
- [x] add to the drawer
- [x] change the link in the footer: https://github.com/openfoodfacts/open-prices-frontend/commit/e734a5f22186c5b1fb84bc44dae2d6bebba02f56
- [ ] serve the frontend app directly prices.openfoodfacts.org : https://github.com/openfoodfacts/open-prices/issues/482
- [x] final frontend config: #906
- [x] redirect any incoming traffic to /app (by removing the prefix): #882

raphodn:
I'm going to take some inspiration from https://hunger.openfoodfacts.org :)

PR #874, opened 2024-09-29
feat(About): new /about page

raphodn:
### What

See #364

Take the content of the current homepage - https://prices.openfoodfacts.org - and move it to a new `/about` page

### Screenshot

image


raphodn:
@raphael0202 I moved about 90% of the current about page
- I removed the section on private proofs
- I did some minor rewording (like renaming some "food prices" to generic "product prices")

raphodn:
I'm not sure it's compatible. The content is now translatable in Crowdin, there can be any language.

Indexing is definitely a topic, but in an other issue ?

raphael0202:
Nice work!
What language do we use by default if the user is not logged-in? The browser-defined language?

raphodn:
> Nice work! What language do we use by default if the user is not logged-in? The browser-defined language?

I believe so ! though this could would need some second eyes because it's a bit old

raphodn:
ok to merge ? so that we can then : 
- have crowdin run the translations
- do another PR to add / fix the links to this page
- do a final PR to remove the ``/app` prefix :star_struck:

raphodn:
French translation done here : #884

PR #883, opened 2024-09-30
refactor(User Dashboard Lists): properly remove Prices and Proof cards after delete

TTalex:
### What
In user dashboard lists:
- Deleting a proof did not visually remove the ProofCard and did not update the number of proofs
- Deleting a price did remove its card, but did not update the number of prices

In both cases, this is properly updated when refreshing the page.

To fix this, I'm forwarding the "delete" event all the way from the dialogs to the UserDashboardPriceList and UseDashboardProofList views.

The views then handle the event and remove the deleted element from their respective lists.

#### Price list before / after change

| Before | After | 
| --- | --- | 
| price-del-before |   price-del-after | 

#### Proof list before / after change

| Before | After | 
| --- | --- | 
| proof-del-before | proof-del-after |

### Notes
I'm not too sure on how this PR handles the edge cases when pagination takes effect. For example, what happens when deleting the last price of a page ?

### DIFF
--- a/src/components/PriceActionMenuButton.vue
+++ b/src/components/PriceActionMenuButton.vue
@@ -48,7 +48,7 @@
     v-if="deleteConfirmationDialog"
     v-model="deleteConfirmationDialog"
     :price="price"
-    @delete="showDeleteSuccessMessage"
+    @delete="handlePriceDelete"
     @close="closeDeleteConfirmationDialog"
   />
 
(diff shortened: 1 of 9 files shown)


TTalex:
Hmm, something's not great... looks like this solution eats the success messages, since they are part of the Cards components.

raphodn:
Honestly this subject is tricky, I remember we spent quite a lot of time on it with @dq18 , without totally satisfying results..

So no need to spend hours on it ;)

TTalex:
> So no need to spend hours on it ;)

Fair, I'll close the PR until we get more user feedback asking for a change :)

We'll probably find a better solution by then.

PR #906, opened 2024-10-01
fix: kill the landing page

raphael0202:


package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "build-staging": "vite build  --base=/app/ --mode preprod",
-    "build-prod": "vite build --base=/app/ --mode prod",
+    "build-staging": "vite build mode preprod",
```
raphodn, line 9: the CI issue is probably because you changed "--mode" to "mode"

raphodn: fix here : https://github.com/openfoodfacts/open-prices-frontend/commit/011d5a1e25419824670fc1956a7098225058a064

raphael0202: wow sorry for that :sweat_smile:

Issue #839, opened 2024-09-23
Dashboard : add extra filters on my proofs


raphodn:
### Story

Some users have many proofs, and want to easily filter through them (in order to find an existing proof for instance).

Currently there is only a filter to hide proofs with at least 1 price.

We could add new filters such as : 
- date (year, month, day)
- type (receipt or price tag or GDPR)
- location (brand, city, country ? slightly tricky)
- ?

serpico:
Currently :

- clicking on the user chip send to the user page
- clicking on the date chip send the user to the date page
- clicking on the location chip send the user to the location page

However there's no effect ( not for me anyway ) when clicking on Price tag chip , is it the same for Receipt chip ?


![ChipToProof](https://github.com/user-attachments/assets/2493c352-0e32-4179-8d55-cffd9292d721)

raphodn:
in the Proof modal, clicking on the "Price chip" should work if you're the price owner. And send you to a `/proofs/id` page.
though this is will probably change/be improved in the coming months, as there is more privacy-friendly features

serpico:
@raphodn how much work would it take to add order like  :

- proof with the greater number of prices first/at the top vs oldest first/at the top
- oldest date first/at the top vs newest date first/at the top

although I could totally wait for the more versatile/powerful search  feature #839

raphodn:
> proof with the greater number of prices first

isn't this already available ? or did you mean proof with least number of prices ?

serpico:
> > proof with the greater number of prices first
> 
> isn't this already available ? or did you mean proof with least number of prices ? i feel this could be better managed with (super fine-grained) filters

Yeah it's available, but once you display the least number on top then you need the option greater number to switch back...

I agree with you about the super-fine grained filters...I mentioned it in my original message, on the PR but I failed to copy that part here.

raphodn:
sorry i had removed your last sentence, i put it back ^^

and sorry but i still don't understand. there is already the option to order by "price_count DESC". you want to be able to order by "price_count ASC" as well ?

raphodn:
> @snowflakes2499
We should allow sorting by proof date, missing prices, and total amount.
Possible fix: Modify the sorting function in this file to support different criteria using state.

In the dashboard it's already possible to order by proof date. you meant asc/desc ?
In the "all proofs" page the default order is by date desc.

|Current filters|Current order|
|-|-|
[...]

PR #907, opened 2024-10-01
refactor(Dashboard): fix Proof count. Cleanup Price & Proof headers

raphodn:
### What

User Prices
- move Dashboard link down
- remove Share button

User proofs
- fix count
- move Dashboard link down

### DIFF
--- a/src/views/UserDashboardPriceList.vue
+++ b/src/views/UserDashboardPriceList.vue
@@ -3,24 +3,15 @@
     {{ $t('UserDashboard.MyPrices') }}
   </h1>
 
-  <v-row>
+  <v-row v-if="!loading">
     <v-col>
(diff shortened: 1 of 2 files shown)


raphodn:
@serpico possibly fixes another bug on the proof count that was displayed

serpico:
@raphodn how much work would it take to add order like  :

- proof with the greater number of prices first/at the top vs oldest first/at the top
- oldest date first/at the top vs newest date first/at the top

although I could totally wait for the more versatile/powerful search  feature #839

raphodn:
ordering is easy :) can you copy-paste your message in #839 ? I renamed the issue to include ordering as well !


but filtering is a bit harder, at least how it works right now is not scalable (realized this in #904)

PR #921, opened 2024-10-03
feat(App): Dark mode support 

### LINKED ISSUE #274: Add a dark mode

TTalex:
### What
- Implemented dark mode 🥷 support using vuetify themes
- For now it only reads user preferences from its browser, there is no button to swap between modes
- Added theme params `header` and `footer` to match the custom colors that were set inline
    - This unlocks classes like `bg-header`
- Also added theme support for PriceChart
- Also added theme support for LeafletMap
- Also added missing OSM attribution in LeafletMap while I was at it

### Screenshot
darkmode

(Light mode still works)
lightmode

### Fixes bug(s)
- Mostly fixes #274


raphodn (approved):
nice :sunglasses:

src/App.vue
```diff
@@ -12,6 +12,17 @@
   </v-app>
 </template>
 
+<script setup>
+import { useTheme } from 'vuetify'
+
+const theme = useTheme()
+
+const prefersDarkScheme = window.matchMedia('(prefers-color-scheme: dark)')
```
raphodn, line 20: easy to add a settings toggle in a following PR ?
I know some users will be asking for it to revert back to the light mode ^^

raphodn: in https://hunger.openfoodfacts.org its interesting because the settings are not "hidden" behind auth

raphodn: it could be a header (or footer) toggle. vs a new settings. or move the settings outside of the auth wall ? :thinking:

src/components/LeafletMap.vue
```diff
@@ -31,6 +32,16 @@ export default {
       required: true
     },
   },
+  setup() {
```
raphodn, line 35: fyi I've currently been using Vue.js with the "options" API. here the setup refers to the "composition" API. I don't know how they both cohabitate ?

src/App.vue
TTalex: I'll do it in this PR, just wanted an insight on where to put the toggle :)

I do not think it should require login, and I don't know if I want to tackle moving the setting page :smile: 

The Footer has a lot of available real estate, so I guess I'll put something there for now ?

src/components/LeafletMap.vue
TTalex: Ah, my bad, I'll check how to properly do this during the weekend.

src/App.vue
raphodn: Footer sounds great !

TTalex:
Now with a button in footer

mode_toggle

It still reads browser preferences until the button was pressed once, then it remembers the user choice and ignores browser configuration. (This is the same behavior as Hunger Games)

src/i18n/locales/fr.json
```diff
@@ -273,6 +273,10 @@
 		"SignIn": "Se connecter",
 		"SignOut": "Se déconnecter"
 	},
+	"Theme": {
```
raphodn, line 276: I'm not 100% sure these translations will actually be taken into account by Crowdin. What I've been doing is only change the en.json. We can leave it like this and test ^^

src/store.js
```diff
@@ -42,6 +43,9 @@ export const useAppStore = defineStore('app', {
     },
     getUserLastCurrencyUsed: (state) => {
       return state.user.last_currency_used
+    },
+    getUserPreferedTheme: (state) => {
```
raphodn, line 47: :+1:

raphodn (approved):
nice v2 !

src/i18n/locales/fr.json
raphodn: Had to resolve a conflict in #929
image

==> Crowdin only looks at en.json

TTalex: Alright, good to know, thanks !

PR #955, opened 2024-10-15
refactor(ProofCard): Only show proof thumb image in User Dashboard Proof List

### LINKED ISSUE #952: We can't read any more from receipts due to poor quality
Following #924

On the web version, the switch to images of size 400px made it impossible to read for most receipts.

I would suggest to:

- display the full version when you're on the price addition form
- in other cases, open a modal when clicking on a proof image to display proof full size.

TTalex:
### What

- Following #924
- Only show proof thumb image in User Dashboard Proof List
- All other pages should display the original proof image, as to not offuscate important details

### Fixes bug(s)
- Fixes #952

### DIFF
--- a/src/components/ProofCard.vue
+++ b/src/components/ProofCard.vue
@@ -7,7 +7,7 @@
     <v-divider v-if="!hideProofHeader" />
 
     <v-card-text>
-      <v-img v-if="proof.file_path" :src="getProofUrl(proof)" :style="'max-height:' + imageHeight" />
+      <v-img v-if="proof.file_path" :src="getProofUrl(proof, showThumbImage)" :style="'max-height:' + imageHeight" />
     </v-card-text>
 
     <v-divider />
@@ -50,6 +50,10 @@ export default {
       type: String,
       default: '100%',
     },
+    showThumbImage: {
+      type: Boolean,
+      default: false,
+    },
   },
   emits: ['proofSelected', 'close'],
   data() {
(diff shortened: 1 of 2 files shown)


src/components/ProofCard.vue
```diff
@@ -63,10 +67,10 @@ export default {
         this.$emit('proofSelected', this.proof)
       }
     },
-    getProofUrl(proof) {
+    getProofUrl(proof, showThumbImage) {
       // return 'https://prices.openfoodfacts.org/img/0002/qU59gK8PQw.400.webp'  // PRICE_TAG
       // return 'https://prices.openfoodfacts.net/img/0001/lZGFga9ZOT.400.webp'  // RECEIPT
-      if (proof.image_thumb_path) {
+      if (proof.image_thumb_path && showThumbImage) {
```
raphodn, line 73: I think to simplify i'd rather have the rule only set here (and not have the 2nd param in getProofUrl)
- avoids using the prop in the template
- avoid confusion between the prop name and the method param name

```suggestion
      if (proof.image_thumb_path && this.showThumbImage) {
```

src/components/ProofCard.vue
```diff
@@ -50,6 +50,10 @@ export default {
       type: String,
       default: '100%',
     },
+    showThumbImage: {
```
raphodn, line 53: the backend field name is `image_thumb_path` so I'd rather have the same order for the prop for clarity :)

```suggestion
    showImageThumb: {
```

raphodn (approved):
2 comments, but good to go afterwards :pray:

src/components/ProofCard.vue
raphodn: and when you think about it, the `getProofUrl` method could also very well work without any param at all (and call this.proof instead !)

TTalex: I did actually get confused by that. So I opted to remove all params from `getProofUrl`

Thanks for the suggestion

raphodn: also did a last small fix and moved getProofUrl to computed (instead of methods)

PR #956, opened 2024-10-15
feat(SignIn): parse and use username from access_token on signin, allows usage of emails to login

### LINKED ISSUE #947: Users should not be able to login with their email (username only)
### What

We should not allow users to login with their email.
We currently only manage the username.
- show an error message if an email is used

### Linked issues

#571

TTalex:
### What
I've noticed that my acces_token always takes the following shape: `<username>__U<uuid>`, even when I use an email as login.

If this is true, maybe parsing this access token, extracting the username and feeding it to the appStore could allow user to still use email as login.

### Fixes bug(s)
- Fixes #947

### Note
This might require some testing, I couldn't find documentation on what the OFF token should look like, so I only have my own experience as data.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -49,8 +49,8 @@ export const useAppStore = defineStore('app', {
     }
   },
   actions: {
-    signIn(username, token) {
-      this.user.username = username
+    signIn(token) {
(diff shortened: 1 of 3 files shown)


raphodn (approved):
awesome !

PR #989, opened 2024-10-27
feat(Header): hide old User settings, move Sign Out to drawer

raphodn:
### What

Header changes (when logged in)
- remove the user menu dropdown
- clicking on the user icon sends to the dashboard

Drawer (when logged in)
- change Dashboard icon (now same as user in the header)
- add Sign Out button at the bottom

### DIFF
--- a/src/components/Header.vue
+++ b/src/components/Header.vue
@@ -23,34 +23,30 @@
       </v-btn>
     </template>
     <template v-else #append>
-      <v-menu scroll-strategy="close">
-        <template #activator="{ props }">
-          <v-btn v-if="!$vuetify.display.smAndUp" v-bind="props" icon="mdi-account-circle" />
-          <v-btn v-else v-bind="props" class="text-lowercase" prepend-icon="mdi-account-circle">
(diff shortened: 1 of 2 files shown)


TTalex:
@raphodn I kinda miss the ability to sign out by clicking on the user icon in the header. 
I feel like it has become a pretty standardized way of signing out, even when there is another menu (GitHub is a good example since it has both, yet you still sign out on the top right)

raphodn:
How often do you sign out ?
Its usually "very easy" to sign in, and "very hard" to sign out :)

Motivation was
- mimick hunger games
- 1-click to go to your dashboard (instead of 2 with the intermediate dropdown)

raphodn:
You're right though that on apps where there are a lot of User-related pages (for instance Github), the websites have a top right dropdown with all these links.

Maybe we'll have to reintroduce it (the dropdown) if we add more authenticed-only pages ?

TTalex:
Yeah, although that's pretty much the job of the user dashboard, so I do not see it being useful soon.

PR #1025, opened 2024-11-09
feat(ContributionAssistant): new experiment, a contribution assistant using gemini

TTalex:
### What
- UI for an automatic parser of labels using Gemini AI
- See the source project for more details https://github.com/TTalex/open-prices-contribution-tools

Added features compared to the source:
- Integration with Open Prices components
- Added barcode detection, users can freely choose between modes (category vs barcode)
   - This makes it a bit more versatile, and no longer limited to fruits and vegetables only if barcodes are visible 
- Fixed display on small devices
- Cropping now also works on mobile, but the UX is still pretty awful, so mouse/keyboard is still recommended
- Removed login step, since it can use the global one
- All prices are added under a single proof, containing the full image, instead of many cropped proofs

### Demo video / tutorial

https://github.com/user-attachments/assets/c1793ee8-ba64-40f3-93c5-161cb164e614

### Requires backend update
- https://github.com/openfoodfacts/open-prices/pull/557

### DIFF
--- a/src/components/ContributionAssistantDateForm.vue
+++ b/src/components/ContributionAssistantDateForm.vue
@@ -1,0 +1,33 @@
+<template>
+  <v-form>
+    <h3 class="required mb-1">
+      2. Select the date of the pictures
+    </h3>
+    <v-text-field
+      v-model="date"
+      label="Date"
+      type="date"
+      :max="currentDate"
+      :onChange="onDateChange"
+      hide-details="auto"
+    />
+  </v-form>
+</template>
+  
+<script>
+  export default {
+    emits: ['date'],
+    data() {
+      return {
+        currentDate: new Date().toISOString().substring(0, 10),
+        date: new Date().toISOString().substring(0, 10)
+      }
+    },
+    methods: {
+      onDateChange() {
+        this.$emit("date", this.date)
+      },
+    }
+  }
+  </script>
+  
--- a/src/components/ContributionAssistantPriceFormCard.vue
+++ b/src/components/ContributionAssistantPriceFormCard.vue
@@ -1,0 +1,84 @@
+<!-- eslint-disable vue/no-mutating-props -->
+<template>
+  <v-card
+    class="mb-4"
+    title="Label"
+    prepend-icon="mdi-tag-outline"
+    height="100%"
+    style="border: 1px solid transparent"
+    :color="productPriceForm.processed ? 'success' : ''"
+  >
+    <v-divider />
+    <v-img
+      height="200px"
+      :src="productPriceForm.proofImage"
+      contain
+    />
+    <v-card-text>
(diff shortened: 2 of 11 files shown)


src/components/ContributionAssistantPriceFormCard.vue
```diff
@@ -0,0 +1,84 @@
...
+    :color="productPriceForm.processed ? 'success' : ''"
+  >
+    <v-divider />
+    <v-img
+      height="200px"
+      :src="productPriceForm.proofImage"
+      contain
+    />
+    <v-card-text>
+      <ProductInputRow :productForm="productPriceForm" :disableInitWhenSwitchingModes="true" @filled="productFormFilled = $event" />
```
raphodn, line 18: maybe the mode init could be done beforehand ?
asking if its a photo of barcodes, or categories ?

src/components/ContributionAssistantDateForm.vue
raphodn, line 1: we could build a dedicated Date component ?

raphodn (approved):
genai ftw!
approving so that we can merge quickly (when the backend is ready)

src/components/ContributionAssistantDateForm.vue
raphodn: actually, re-watching your video, why not use the ProofMetadataInputRow ? because the currency is important as well.

src/components/ProofImageInputRow.vue
raphodn, line 1: I pushed a commit to add this new prop: https://github.com/openfoodfacts/open-prices-frontend/commit/3ece9a0d0240a0c4af7c67ab86046bb1002a4968 (so you'll need to rebase before merging :sweat_smile: )

src/components/ContributionAssistantDateForm.vue
TTalex: Good idea !

src/components/ProofImageInputRow.vue
TTalex: I think I did it ? Without creating a big mess ?

raphodn: oh yeah i guess there were some merge conflicts, sorry...

raphodn: when we'll be ready to merge I don't mind doing the last rebase to cleanup (the translations for instance)

raphodn:
did a few small changes here : https://github.com/openfoodfacts/open-prices-frontend/commit/a2ae48d6890770706a60f7923c9b804152bdb8e9
- new icon
- spacing, button color, some existing translations, wider labels
- redirect to dashboard at the end (like other price addition forms)

TTalex:
My idea behind not redirecting users after price addition was to let them add another proof without having to select the date and store again, by going back to the previous step.

I believe it might be better to have a new tab that confirms the successful addition of prices and suggest a next step with links, either to the user dashboard or to go back to the proof page (and maybe clear the image ?)

raphodn:
ah sorry, feel free to revert the "redirect" change :bow:

PR #1035, opened 2024-11-13
refactor(API): send the page url on every POST

raphodn:
### What

Following #1034 & #1025.
A bit of a hack, where we send the user's current page URL on each POST.

### Why

To have some stats on where the proofs & prices are created (as we have multiple forms now).


raphodn:
fyi @TTalex so that we have some usage stats on the Contribution Assistant (and other "experiments").

PR #1039, opened 2024-11-14
refactor(ContributionAssistant): More UX tweaks following feedback

TTalex:
### What
- Display detected barcode below product suggestion, to compare
- Show more cols on larger screens
- Hide "price per" input when barcode is selected
- Allow for price removal at the cleanup step, with a trashcan icon
- Add a 4th step: Summary, with progress bars tracking proof and price uploads and navigation buttons to user profile, added proof or to go back to the image selection step

### Screenshots
tweaks
summary

### Part of 
- #1025

### DIFF
--- a/src/components/ContributionAssistantPriceFormCard.vue
+++ b/src/components/ContributionAssistantPriceFormCard.vue
@@ -16,6 +16,14 @@
     />
     <v-card-text>
       <ProductInputRow :productForm="productPriceForm" :disableInitWhenSwitchingModes="true" @filled="productFormFilled = $event" />
+      <v-alert
+        v-if="productPriceForm.mode === 'barcode'"
+        class="mb-2"
+        type="info"
+        variant="plain"
+      >
+        Detected barcode: {{ productPriceForm.detected_product_code }}
+      </v-alert>
       <v-row>
         <v-col>
           <h3 class="required mb-1">
@@ -60,7 +68,8 @@ export default {
         price_without_discount: null,
         currency: null,
         proofImage: null,
-        processed: null
+        processed: null,
+        detected_product_code: null
       })
     },
   },
--- a/src/components/ContributionAssistantPriceFormCard.vue
+++ b/src/components/ContributionAssistantPriceFormCard.vue
@@ -3,19 +3,13 @@
   <v-card
     class="mb-4"
     height="100%"
+    title="Label"
+    prepend-icon="mdi-tag-outline"
     style="border: 1px solid transparent"
   >
-    <v-card-item>
-      <template #prepend>
-        <v-icon icon="mdi-tag-outline" />
-      </template>
-      <h2 class="text-h6">
-        Label
-      </h2>
-      <template #append>
-        <v-btn icon="mdi-delete" @click="removePrice()" />
-      </template>
-    </v-card-item>
+    <template #append>
+      <v-icon icon="mdi-delete" color="error" @click="removePrice()" />
+    </template>
     <v-divider />
     <v-img
       height="200px"
--- a/src/components/ContributionAssistantPriceFormCard.vue
+++ b/src/components/ContributionAssistantPriceFormCard.vue
@@ -2,12 +2,20 @@
 <template>
   <v-card
     class="mb-4"
-    title="Label"
-    prepend-icon="mdi-tag-outline"
     height="100%"
     style="border: 1px solid transparent"
-    :color="productPriceForm.processed ? 'success' : ''"
   >
+    <v-card-item>
+      <template #prepend>
+        <v-icon icon="mdi-tag-outline" />
+      </template>
+      <h2 class="text-h6">
+        Label
+      </h2>
+      <template #append>
+        <v-btn icon="mdi-delete" @click="removePrice()" />
+      </template>
+    </v-card-item>
     <v-divider />
     <v-img
       height="200px"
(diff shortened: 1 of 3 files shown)


src/components/ContributionAssistantPriceFormCard.vue
```diff
@@ -2,12 +2,20 @@
...
   <v-card
     class="mb-4"
-    title="Label"
-    prepend-icon="mdi-tag-outline"
     height="100%"
     style="border: 1px solid transparent"
-    :color="productPriceForm.processed ? 'success' : ''"
   >
+    <v-card-item>
+      <template #prepend>
```
raphodn, line 9: looking at a similar UX in `AddPriceMultiple.ProductPriceDetails.NewPrice` i think we can simplify this bit, and keep the label & prepend-icon in the `<v-card>` definition, no ?

src/components/ContributionAssistantPriceFormCard.vue
```diff
@@ -60,25 +76,26 @@ export default {
...
-    ],
-   }
-  },
-  computed: {
-  },
-  mounted() {
+    return {
+      productFormFilled: false,
+      pricePriceFormFilled: false,
+      categoryPricePerList: [
```
raphodn, line 89: note to self: refactor this somehow (used in 3 files)

src/views/ContributionAssistant.vue
```diff
@@ -10,6 +10,9 @@
       <v-tab value="Cleanup" :disabled="!productPriceForms.length">
         3. Cleanup
       </v-tab>
+      <v-tab value="Summary" :disabled="!(productPriceForms.length && (addPricesLoading || productPriceForms.length == numberOfPricesAdded))">
```
raphodn, line 13: if the disabled rules are getting too complicated, we can move them to a `method`

raphodn: also, when on the Summary tab, the other tabs should be disabled, no ? to avoid duplicate actions. or do you want to allow users to go back ? add `<v-overlay>` then ?

src/views/ContributionAssistant.vue
```diff
@@ -98,6 +102,45 @@
...
+                v-if="!addProofLoading"
+                v-model="numberOfPricesAdded"
+                :max="productPriceForms.length"
+                :color="productPriceForms.length == numberOfPricesAdded ? 'success' : 'info'"
+                height="25"
+                stripped
+              >
+                <strong>{{ numberOfPricesAdded }} / {{ productPriceForms.length }} prices added</strong>
+              </v-progress-linear>
+              <v-btn to="/dashboard" class="mt-4" :aria-label="$t('Common.Dashboard')" :disabled="productPriceForms.length != numberOfPricesAdded">
```
raphodn, line 131: these 3 disabled could be transformed into 1 computed value "done" for instance ? + the above progress. 
or simply replace with `addProofLoading` ?
when would numberOfPricesAdded be different from productPriceForms.length ? (sure if the API bugs at that very moment, but unlikely, and maybe overkill to manage errors right now ?)

raphodn (approved):

src/components/ContributionAssistantPriceFormCard.vue
```diff
@@ -16,12 +24,20 @@
     />
     <v-card-text>
       <ProductInputRow :productForm="productPriceForm" :disableInitWhenSwitchingModes="true" @filled="productFormFilled = $event" />
+      <v-alert
```
raphodn, line 27: last bit missing, is to set the mode as barcode if a product_code is detected ?? or is there many false-positives ?

src/views/ContributionAssistant.vue
TTalex: productPriceForms.length tracks the total number of prices to push to API
numberOfPricesAdded tracks the number of prices handled by API

Since prices are added one by one, the values are different during the upload process, which can take a few seconds for many prices.

addProofLoading only tracks the proof API, when it's done, there still are a lot of prices to be added.

Agreed that it should be refactored as a "done" computed value !

src/components/ContributionAssistantPriceFormCard.vue
TTalex: Sorry, didn't understand this comment.

But maybe this context helps: gemini always tries its best to but a value to all fields. So its reply almost always contains a barcode, even when no barcodes were in the image (it likes to put the price in the barcode field for these cases)

Related fun fact: I've noticed it also does something similar with origin countries that are not in its enum set, replacing it with countries that are close geographically or culturally. For example, if its set includes CHINA and BRAZIL, and the read origin is CHILE, it tends to reply with BRAZIL, instead of the closer word CHINA, which is surprising (and pretty cool).

src/components/ContributionAssistantPriceFormCard.vue
TTalex: Definitely ! Thanks for the reference.

For some reason, when I tried that yesterday, the delete icon wouldn't align correctly. Could be because I was trying to use a button instead of a clickable v-icon.

src/components/ContributionAssistantPriceFormCard.vue
TTalex: So the suggestion is to have "barcode" selected by default instead of "category" ?

I wonder if users would also get lost for crops without barcodes, with gemini putting a random number as barcode, and users not knowing that the category is hidden behind the other mode.

TTalex:
I ended up following the advice of moving the proof upload process as the first step.
tweaks

This is more in-lines with other processes of Open Prices (reuses the ProofInputRow components)

It breaks the idea of handling multiple proofs for the same location, but feedback seems to show that this is not how the tool is used.

Moreover, with the multiple-proof upload page in construction, I guess the process will be to use this feature to add many proofs for a store at once, then use the contribution assistant on the existing proofs.

Using recent proofs skips the compression and proof upload process, which slows down the assistant quite a lot.

src/components/ContributionAssistantPriceFormCard.vue
raphodn: I need to take a closer look in the gemini response. Or maybe couple gemini with a classifier ? That would tell us if the image is a barcode or a category ??

TTalex: There's probably a way of modifying the prompt, asking specifically for EAN13 barcodes or empty strings. It might even lead to better parsing results !

TTalex: See #1042

PR #1042, opened 2024-11-17
refactor(ContributionAssistant): use barcode mode when codes look credible

TTalex:
### What
- Simple PR in reference to @raphodn suggestion in #1039
- If gemini detects something that looks like a barcode (more than 10 characters), it's better to suggest the barcode mode instead of the category one


raphodn (approved):
cool !

PR #1068, opened 2024-11-30
refactor(UI): right-align all submit buttons. change color of Done buttons

raphodn:
### What

After some extensive price-adding with Justin this weekend, first of a few PRs to sightly improve the UI.

Submit buttons : 
- always right-align (people hold their phones with their right hand... though I'm left-handed ^^)
- avoid confusion between Submit & Done buttons

### Screenshot

|Page|Before|After|
|---|---|---|
|Add multiple prices|image|image|


raphodn:
A nice-to-have idea could also be to allow the user to set which side (left or right) are the submit buttons ^^

Issue #1085, opened 2024-12-04
Revamp the price addition workflow


raphodn:
### Story

After some extensive scanning with Justin this weekend, I'm thinking of many small changes to apply to the main "proof+price(s) addition" workflow

### Todo

- [x] when adding a new price, automatically scroll down to the form: https://github.com/openfoodfacts/open-prices-frontend/pull/1072
- [x] right-align submit buttons: https://github.com/openfoodfacts/open-prices-frontend/pull/1068
- [ ] image upload: allow drag-and-drop: https://github.com/openfoodfacts/open-prices-frontend/issues/1084
- [ ] Price: move discount next to label
- [ ] Product: seperate product type & details ? move type next to label ? 
- [ ] required fields: turn the * from red to green when the field is valid ?
- [ ] Proof image: hide the "Select a proof" line
- [ ] Proof location: hide the "Select a location" line
- [ ] ...

### Related issues

- https://github.com/openfoodfacts/open-prices-frontend/issues/930
- https://github.com/openfoodfacts/open-prices-frontend/issues/974

thibaultmol:
Another thing I was thinking of: I actually would love to be able to upload proofs at once. just dump them all in so I no longer need to worry about those files on my pc/phone.

In the current flow I have to remember which images i've already uploaded and which I haven't. to the point where i just open multiple tabs and upload all the images. so that I don't have to worry about it.

serpico:
did you try https://prices.openfoodfacts.org/proofs/add/multiple ? ( it's in the side menu > Experiments > Add multiple proofs )

raphodn:
@serpico ok so you're not going to be happy with the layout i guess ^^
- it's only half done (i need to take some time for the "normal" form fields)
- and think of an expert mode to save some clicks (i know, the new UI adds some clicks, I tried it quite a bit this weekend)

thibaultmol:
I like the change of 'find the product' instead of whatever barcode it was before. 
i really like how the barcode is already prefilled in the 'type the barcode' field, not sure if that was the case before

one thing i would like to ask: could we have the cropped image of the product, very large, above the 'type the barcode' field.
![Image](https://github.com/user-attachments/assets/47c5feed-6eb5-4bd5-8d77-7e0d8945f868)

right above this i would like to see the price time image massive, so i can easily read it.

[...]

PR #1108, opened 2024-12-09
refactor(Price add): On barcode manual input, show result confirmation before send

raphodn:
### What

Following #1102 where we merged the barcode scan & manual input in 1 single dialog.

Changes:
- minimum dialog height
- show product result on search + user needs to select the product for confirmation

Following feedback in #1085

### Screenshot

image

### DIFF
--- a/src/components/BarcodeManualInputDialog.vue
+++ b/src/components/BarcodeManualInputDialog.vue
@@ -32,7 +32,7 @@
 <script>
 export default {
   props: {
-    preFillValue: {
+    barcodeInputPrefillValue: {
       type: String,
       default: ''
     }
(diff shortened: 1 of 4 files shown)


TTalex:
Pretty cool !

Small detail, this seems to break pre-filling the barcode value. At a glance, it looks like there is a mismatch in variables names, but I haven't tested if that's the only issue.

- `barcodeInputPrefillValue` in ProductInputRow parent 
- `barcodeManualInputPrefillValue` in props declaration
- `this.preFillValue` in mounted.

raphodn:
> Small detail, this seems to break pre-filling the barcode value.

ahh you're right, i renamed the props but not quite everywhere :eyes: 
pushing a fix

PR #1124, opened 2024-12-12
feat(ContributionAssistant): Add support for automatic label detection

TTalex:
### What
- Following the new ml detection of price tags bounding boxes https://github.com/openfoodfacts/open-prices/pull/614
- This PR uses this new result to speed up the cropping step

### Screenshot

After adding a proof, an info message is shown.

Once the bounding boxes are retrieved, images are automatically cropped. Users can still remove or add crops.

The assistant tries to fetch bounding boxes twice (using single proof api).
- The first time right after uploading the proof (useful for proof already uploaded beforehand)
- If that fails, a second fetch is done 5 seconds later, giving time for ml to execute

If both tries fail, the following message is shown to the user

### DIFF
--- a/src/components/ContributionAssistantDrawCanvas.vue
+++ b/src/components/ContributionAssistantDrawCanvas.vue
@@ -18,6 +18,10 @@
       image: {
         type: Image,
         default: null
+      },
+      seedCrops: {
+        type: Array,
+        default: null
       }
     },
     emits: ['croppedImages', 'loaded'],
@@ -30,6 +34,13 @@
         rectangles: []
       }
     },
+    watch: {
+      seedCrops() {
+        if (this.seedCrops) {
+          this.init(true)
+        }
+      }
+    },
     mounted() {
       if (this.image.complete) {
         this.init()
@@ -38,7 +49,7 @@
       }
     },
     methods: {
-      init() {
+      init(keepRectangles=false) {
         const canvas = this.$refs.canvas
         const ctx = canvas.getContext("2d")
         canvas.style.width = "100%"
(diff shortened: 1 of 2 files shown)


src/components/ContributionAssistantDrawCanvas.vue
```diff
@@ -18,6 +18,10 @@
       image: {
         type: Image,
         default: null
+      },
+      seedCrops: {
```
raphodn, line 22: could we find another name ? when I read this I imagine myself in an farmer's crop/field ^^
the backend sends `bounding_box` arrays. 
- maybe replace `rectangles` by `boundingBoxes`
- and `seed` by `initFromServer` or something like this ?

raphodn: honestly it's not important, we can refactor later if you feel like/find a better name ^^

src/views/ContributionAssistant.vue
```diff
@@ -197,7 +210,7 @@ export default {
     reloadPage(){
       window.location.reload()
     },
-    setProof(event) {
+    async setProof(event) {
```
raphodn, line 213: we don't have much/any async/await in the code elsewhere.
just saying it'll be slightly harder to maintain if we start having both in the code.
can wait a further refactor

src/views/ContributionAssistant.vue
```diff
@@ -208,6 +221,28 @@ export default {
...
+        // If that also fails, user will have to click the button to retry
+      }
+    },
+    async loadPredictions(proofId) {
+      this.predictionLoading = true
+      const proof = await api.getProofById(proofId)
+      if (proof.predictions && proof.predictions.length) {
+        for (let prediction of proof.predictions) {
+          if (prediction.type === "OBJECT_DETECTION") {
+            this.seedCrops = prediction.data.objects.map(predictionObject => {
```
raphodn, line 239:
```suggestion
            this.seedCrops = prediction.data.objects.map(predictionObject => predictionObject.bounding_box)
```

raphodn (approved):

src/views/ContributionAssistant.vue
```diff
@@ -26,13 +26,24 @@
       </v-tabs-window-item>
       <v-tabs-window-item value="Crop">
         <v-container>
+          <v-alert v-if="drawCanvasLoaded && !seedCrops.length && !predictionLoading" type="info" variant="outlined" icon="mdi-alert">
+            No labels could be automatically detected. Please manually draw squares around the labels or press the button below to try again.
+            <br />
+            <v-btn @click="loadPredictions(proofForm.id)">
+              Automatically find labels
```
raphodn, line 33: we'll need to move this to translations in a future PR

PR #1125, opened 2024-12-13
fix(Price add): fix barcode type prefill feature (other typo). ref #1108

TTalex:
### What
- See #1108 
- Completes commit https://github.com/openfoodfacts/open-prices-frontend/commit/8af89a7183d6a8cd3059f12bdfc4a8aa072717a3


raphodn:
I think I fixed this last night ? 😅

PR #1129, opened 2024-12-14
refactor(ContributionAssistant): var names, translations, avoid async

TTalex:
### What

- No changes in features
- Implemented feedback from https://github.com/openfoodfacts/open-prices-frontend/pull/1124
- Rewrote most variable and methods names (no more "crop")
- Removed all raw text and added translations
- Rewrote methods to avoid using async/await
- Added footer to extracted labels list (used to be called cropped images) with bounding boxes sources and delete button on bottom right
- Some better error handling

### Screenshots

|Step|Image|
|---|---|
|Step 2|image|
|Step 3|image|


src/components/ContributionAssistantLabelList.vue
```diff
@@ -0,0 +1,57 @@
+<template>
```
raphodn, line 1: nice ! and I'll do the same in the main price workflow
- re-add a divider at the end to separate the form and the actions
- move the "Trash" button in the footer instead of the top-left

raphodn: done here : https://github.com/openfoodfacts/open-prices-frontend/pull/1129/commits/5e9264f4b38335857490707dd517d884b4adf9aa

raphodn (approved):

src/components/ContributionAssistantLabelList.vue
raphodn: Also wondering if we should create a Label component instead of a LabelList. But it can wait :)

PR #1166, opened 2024-12-23
refactor(ContributionAssistant): Full integration with price tag data

TTalex:
### What
- The contribution assistant now reads (as input), creates (after drawing) and updates price tags (after price creation)
- When no new price tags are drawn, this significantly improves the speed of the contribution assistant for existing proofs, since it no longer requires waiting for gemini. If new price tags are drawn, we still have to wait for backend processing.
- It also supports skipping the proof upload step when a query parameter contains proofs_ids to handle, e.g: `/experiments/contribution-assistant?proofs=[1512,1513]`
- After uploading the prices, users can press the new "Next Proof" button to go back and handle the next proof_id in params.

### Demo
Example, starting from a not yet implemented multiple proof upload CTA
https://github.com/user-attachments/assets/517d535f-8283-42f3-957d-94338af3d4c3

Another example, showing how the assistant can be used to add price tags manually (the price validator assistant is used to show an increasing number of price tags, before and after) 
https://github.com/user-attachments/assets/14a20851-e269-44fc-b5dd-c6278c7ef726

Note that currently, it only supports adding new price tags, and does not handle removing them.

### DIFF
--- a/src/views/ContributionAssistant.vue
+++ b/src/views/ContributionAssistant.vue
@@ -125,6 +125,7 @@
       </v-tabs-window-item>
     </v-tabs-window>
   </v-container>
+
   <v-snackbar
     v-model="labelProcessingErrorMessage"
     color="error"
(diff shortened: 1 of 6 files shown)


src/views/ContributionAssistant.vue
```diff
@@ -209,84 +209,133 @@ export default {
       // Summary tab should be enabled when there are product prices to be added and the add prices process is either running or done
       const enableSummaryTab = this.productPriceForms.length && (this.addPricesLoading || this.allDone)
       return !enableSummaryTab
+    },
+    proofsIdsFromQueryParam() {
+      return JSON.parse(this.$router.currentRoute.value.query.proofs)
```
raphodn, line 214: could we use `proof_ids=id1,id2` ?
- rename proofs to proof_ids
- use comma-seperated values

raphodn (approved):

src/views/ContributionAssistant.vue
raphodn: an alternative could also be `proof_id=id1&next_proof_id=id2`

TTalex: Sure, I spend a bit of time looking for a standardized way of passing arrays in url params. It's quite surprising that there isn't. I went with this one for ease of use using JSON, but I do not like it either 😅

teolemon: proof_ids=id1,id2 has my preference, we already use it for product lists in ProductOpener

raphodn: changed here: https://github.com/openfoodfacts/open-prices-frontend/pull/1166/commits/07711e7b2ea99e31dd543ef85d7c0bac1290d3dd

PR #1188, opened 2024-12-29
refactor(ContributionAssistant): More pricetags integrations

TTalex:
### What
Following @raphael0202 feedback  (translated, simplified):

- [x] 1. proof_ids=ID1,ID2 parameter is not working

Fixed in https://github.com/openfoodfacts/open-prices-frontend/commit/012b83ff4a07433df436b6a5f4e36c4e21e9927c

- [x] 2. Pricetags status and price id are ignored. "3. Cleanup" phase should ignore (grey out) price tags with non-null status

Status and price_id are no longer ignored, Cleanup step now contains a second list of prices, below the first one, with unmodifiable (disabled) already added prices. These prices are ignored in further steps. (Note that the content is still showing Gemini predictions, now the actual created price, which might be a bit misleading)
Capture

- [x] 3. Newly created bounding boxes sometime give API error "Bounding box values should be in the format [y_min, x_min, y_max, x_max]."

Could be happening when max values were greater than min values, e.g. a bouding box drawn from bottom to top. Fixed.

- [ ] 4. Newly created prices tags are sent after pressing the "Send labels" button, we should send them as we go, to give time for gemini process

Not implemented for now. I want to avoid the creation of invalid price tags because of user errors while drawing boxes. Moreover, the price tag creation API takes up to 3 seconds to reply, maybe because it waits for prediction results. Waiting for this breaks the fluidity of drawing successive boxes.

- [x] 5. During "Cleanup", unreadable price tags should be identifiable

The cleanup phase now uses the same buttons as the Price Validator Assistant, and updates price tag status depending on user choice.
Capture

- [x] 6. During "Labels extractions", we could use different colors depending on price tag status (linked to a price, automatic, manual, unreadable ... )

Boxes inside the canvas now have different colors and a legend.
Sans titre

- [ ] 7. In "Summary", we should suggest a new proof to handle

UI is done, but the API call isn't correct, currently it fetches all proofs since the proof api doesn't allow for this kind of filtering (proofs with price tags to validate).
Capture

- [ ] 8. In "Labels extraction", we should show existing prices associated to the proof (in case the assistant wasn't used at first)

Existing price tags are shown. However, existing prices without a price tag association aren't shown.

- [ ] 9. In "Labels extraction", we should be able to edit and existing bounding box, instead of removing and adding a new one (using PATCH in API)

 Not really sure how to do this UI-wise, ideally this would mean resizing an existing box, but that is not trivial to implement. Maybe add an "edit" button that allows for redraw and replacement ?

- [x] 10. Empty bounding boxes can still be created

Should be fixed, boxes smaller than a few pixels are ignored.

- [x] 11. Sometimes barcodes detected by Gemini contain a letter, triggering an alert error.

Should be fixed, everything that is not a number is stripped from product_code.

### DIFF
--- a/src/components/ContributionAssistantDrawCanvas.vue
+++ b/src/components/ContributionAssistantDrawCanvas.vue
@@ -37,7 +37,7 @@
     watch: {
       boundingBoxesFromServer() {
         if (this.boundingBoxesFromServer.length) {
-          this.initCanvas(true)
+          this.initCanvas(false)
         }
       }
     },
@@ -74,14 +74,15 @@
           this.boundingBoxes = [] // reset boundingBoxes
         }
         if (this.boundingBoxesFromServer) {
-          this.boundingBoxes = this.boundingBoxes.concat(this.boundingBoxesFromServer.map(({boundingBox, id}) => {
+          this.boundingBoxes = this.boundingBoxes.concat(this.boundingBoxesFromServer.map(({boundingBox, id, status}) => {
             return {
               startY: boundingBox[0] * this.image.height,
               startX: boundingBox[1] * this.image.width,
               endY: boundingBox[2] * this.image.height,
               endX: boundingBox[3] * this.image.width,
               boundingSource: this.$t('ContributionAssistant.AutomaticBoundingBoxSource'),
-              id: id
+              id: id,
+              status: status
             }
           }))
           this.extractLabels()
@@ -98,7 +99,7 @@
         }
         this.startX = event.offsetX / this.scale
         this.startY = event.offsetY / this.scale
-        this.isDrawing = true;
+        this.isDrawing = true
       },
       drawContent(event) {
         if (event.type == "touchmove") {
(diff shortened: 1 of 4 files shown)


raphodn (approved):

raphael0202:
about

> Existing price tags are shown. However, existing prices without a price tag association aren't shown.

and

> (Note that the content is still showing Gemini predictions, now the actual created price, which might be a bit misleading)

If we fetch all prices associated with the proof during step 2, we could fix both of these issues:

- we would be able to display validated price information from price tags that are associated with a price
- we could show prices without a price tag association (and link this price to an existing price tag)

raphael0202:
It works really well :tada:!

A small bug I found in staging:
To know the source (`automatic` vs manual), we should look at the `created_by` attribute of the price tag. It's null if it's generated automatically, and contains the username of the creator otherwise. If I create a price tag manually, then reload the page, the new created price tag is marked as "automatic".

PR #1204, opened 2024-12-31
refactor(Assistants): barcode pre-cleaning rules to improve matching

raphodn:
### What

Some quick rules to cleanup the barcode string returned by the prediction.
- keep only digits
- remove leading 0s
- special rule for Carrefour (linked to https://github.com/openfoodfacts/open-prices/pull/660)
- also, valid barcodes start at 8 (EAN8)

### Why

To increase product matching

### Screenshot

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -65,9 +65,7 @@ function isValidBarcode(value) {
 }
 
 function cleanBarcode(value) {
-  // remove spaces
-  value = value.replace(/\s/g, '')
-  // keep only digits
+  // keep only digits (remove letters, spaces, special characters)
   value = value.replace(/\D/g, '')
   // special case: 22 digits could be Carrefour
   // 182492/3119789831280/051 -> 1824923119789831280051 -> 3119789831280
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -64,6 +64,21 @@ function isValidBarcode(value) {
   return ((10 - (result % 10)) % 10) === parseInt(paddedValue.charAt(13), 10)
 }
 
+function cleanBarcode(value) {
+  // remove spaces
+  value = value.replace(/\s/g, '')
```
TTalex, line 69: This shouldn't be required, the line below removes everything that is not a digit, including spaces.

raphael0202:
If we want to be compatible with Product Opener, barcodes should be normalized by the way:
https://openfoodfacts.github.io/openfoodfacts-server/api/ref-barcode-normalization/

Some leading 0 should be kept.

raphodn:
> Some leading 0 should be kept.

reverted in https://github.com/openfoodfacts/open-prices-frontend/commit/7a7e48f06dfe42f4e1519fd0b7768b18b474b2c6 :ok_hand:

PR #1224, opened 2025-01-03
refactor(Contribution Assistant): cleanup: small improvements for prices already added

raphodn:
### What

Contribution assistant: small improvements on the cleanup step
- hide the price tag actions for prices already added
  (a better alternative would be to display only the image + price card (not the price form))
- add a bit of padding above the title

### Screenshot

|Before|After|
|---|---|
|image|image|


TTalex (approved):
Nice.

Realistically, we should show the previously added price here instead of a disabled edition Card, but that probably requires more API calls and complexity.

PR #1275, opened 2025-01-15
refactor(Contribution assistant): replace tabs with a stepper

raphodn:
### What

We recently introduced a stepper at the top of the "Proof add multiple" & "Price add multiple" pages.
The only remaining page was the Contribution Assistant, that already had tabs, but that we could homogenize and use steps instead :)

I changed a bit the wording to help fit the stepper on smaller screens (though the CA is better to use on desktop anyway...)

The `v-container` were removed so there some horizontal pixel gains !

### Screenshot

||Before|After|
|---|---|---|
|Desktop|image|image|
|Mobile|image|image|


src/views/ContributionAssistant.vue
```diff
@@ -1,174 +1,168 @@
...
+          <v-btn color="primary" :block="!$vuetify.display.smAndUp" :to="'/proofs/' + proofObject.id" :disabled="!allDone">
+            {{ $t('ContributionAssistant.GoToProof') }}
+          </v-btn>
+        </v-col>
+        <v-col>
+          <v-btn color="primary" :block="!$vuetify.display.smAndUp" :disabled="!allDone" @click="reloadPage">
+            {{ $t('ContributionAssistant.AddNewProof') }}
+          </v-btn>
+        </v-col>
+        <v-col v-if="proofIdsFromQueryParam.length > 1">
```
TTalex, line 141: Not something added with this PR, but this breaks when no query params are set
`proofIdsFromQueryParam` is null, giving a type error

src/components/ProofUploadCard.vue
```diff
@@ -28,10 +28,10 @@
       <v-progress-linear
         v-model="proofObjectList.length"
         :max="proofImageList.length"
-        :color="proofImageList.length == proofObjectList.length ? 'success' : 'info'"
+        :color="proofImageList.length === proofObjectList.length ? 'success' : 'info'"
         height="25"
         :indeterminate="proofObjectList.length ? false : true"
-        striped
+        :striped="roofImageList.length !== proofObjectList.length"
```
TTalex, line 34: Missing a "p" here

src/views/ContributionAssistant.vue
raphodn: thanks, fixed !

src/components/ProofUploadCard.vue
raphodn: thanks, fixed !

PR #1280, opened 2025-01-18
feat(SignIn): proper redirect when sign-in is required

TTalex:
### What
Currently:
- When users try to access a page that requires login, they are redirected to the sign-in page
- After signing in, they are redirected to their dashboard

With this PR:
- We keep track of the page they wanted to access
- After signing in, they are redirected to said page
- Users that pressed the regular sign in button are still sent to their dashboard

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -61,7 +61,9 @@ const router = createRouter({
 
 /**
  * On each page change, check if it needs authentication.
- * If required, but the user is not authenticated (token unknown), then redirect to 'sign-in'
+ * If required, but the user is not authenticated (token unknown):
+ * - then redirect to 'sign-in'
+ * - the initial url is passed as query parameter ?next=, in order to redirect back after login
  */
  router.beforeEach(async (to, from, next) => {
   const store = useAppStore()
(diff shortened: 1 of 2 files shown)


src/router.js
```diff
@@ -70,7 +70,7 @@ const router = createRouter({
     await localeManager.changeLanguage(locale)
   }
   if (to.meta.requiresAuth && !store.user.token) {
-    return next({ name: 'sign-in' })
+    return next({ name: 'sign-in', query: { redirect: to.fullPath } })
```
raphodn, line 73: I'm going to rename the parameter to `next` it's what I usually see used

src/views/SignIn.vue
```diff
@@ -105,7 +105,8 @@ export default {
         })
     },
     done() {
-      this.$router.push({ path: '/dashboard', query: { signinSuccess: 'true' } })
+      const path = this.$route.query.redirect || '/dashboard'
```
raphodn, line 108: border case: if the initial path had a query param, then it will disappear after redirect
but that can be done in the futur somehow ^^

raphodn (approved):

PR #1281, opened 2025-01-18
feat(Challenge): Match btn style with multiple proof upload

TTalex:
### What
- Since buttons do the same thing, we might as well make them look the same
- I kept the "Add pictures" denomination, finding it clearer than "Add proof"


raphodn (approved):

src/components/ChallengeTakePicturesCard.vue
```diff
@@ -19,6 +19,7 @@
         color="primary"
         variant="flat"
         :block="!$vuetify.display.smAndUp"
+        prepend-icon="mdi-image-plus"
         to="/proofs/add/multiple"
       >
         {{ $t('Challenge.StepTakePictures.AddPictures') }}
```
raphodn, line 25: reminds me that when you demoed this and I saw "Pictures", I think we should replace elsewhere in the "Images" by "Pictures" (right now we have 3 wordings for the same thing : Proof, Image & Picture I believe)

PR #935, opened 2024-10-05
feat(Challenge Page): A page for community challenges

TTalex:
### What
- See #934 for description, and questions
- This PR is only used to see code diffs for the prototype, it is not ready for merging

### Fixes bug(s)
- Fixes #934

Issue #934, opened 2024-10-05
A page for community challenges 🏆 


TTalex:
### Idea
The idea of periodic challenges emerged during the OFF Days 2024. With a first challenge around milk and milk substitutes.

These kinds of challenges find good success in the OpenStreetMap community, with either [MapRoulette](https://maproulette.org/) or the [monthly projects (link in french)](https://www.openstreetmap.fr/category/contribute/projet-du-mois/) organized by OSM France.

In order to inform contributors and to track progress, we could imagine a dedicated page.

### Proposed solution
I have started working on a prototype to start a discussion on what this page could contain, see #935 

Here are the prototyped features:

- Title and description of the challenge, indicating how to participate
- Reminder of the product categories targeted by this challenge
- A timeline and progress of the challenge, between two dates
- An exemple of a good contribution to the challenge (proof and price)
- General stats : number of prices added during the challenge, number of contributors
- Personalized stats for logged-in users:  number of prices added during the challenge, rank among contributors
- Current top 5 contributors, ordered by number of prices added
- Graph with contributions per day, to track the community efforts
- List of most recent contributions

![Image](https://github.com/user-attachments/assets/097c7e60-ecf2-4435-af10-5fde6db8f06f)

### Questions & remarks

1. Should this be part of the Open Prices App, or included in a different App ?
2. How do we deal with content updates, when a new challenge is created, should such update require edits to the code ?
3. How does this page link to other OFF mediums ? Such as wiki and communication channels ? Should tutoriels like using the app be included on the page ?
4. Should we limit the type of proof for challenge to only accept shelf pictures ? The goal should not be to push people to buy products.
5. Maybe another type of contribution to the challenge should be proofs, flagged as matching the challenge, letting other contributors enter the prices, or even deduce them with OCR.
6. Maybe there could be a goal with a given number of prices to reach, along with the deadline (or replacing it?)
7. What kind of rewards could we give to the best contributors ? Is there a badge system in OFF ? Or should the prize be physical (goodies) ?
8. Right now, the prototype is not working because of missing API points:       
    - getPrices does not allow filtering on specific product categories
    - There is no way of getting a list of users that have contributed to a given category during dates, other than loading all contributions
    - In the same vein, there is no way of getting a total number of contributions per day during dates, other than loading all contributions

raphodn:
I think we need a dedicated time to talk about this. I also feel there are 2 discussions here : 
- how should the challenge's dashboard look like (frontend)
[...]

PR #1284, opened 2025-01-19
refactor(Challenge): improved UI following feedback

TTalex:
### What
* Removed broken number of contributors stat
* Removed Robot emoji from validate prices button, added “[...] with AI” to the title
* Split the main title and the topic (Nutella shouldn't be on the title)
* Moved stat card inside the "quest" cards, instead of below
* Made stat boxes outlined instead of tonal, to avoid confusion with input fields
* Added a mention to Nutella in price stat cards
* Clicking on a picture opens it in a new tab
* Switched timeline to secondary color scheme, to improve readability in light mode
* Rename the page to "Community Challenge" instead of just "Challenge"

### Screenshot
Screenshot 2025-01-19 at 14-38-51 Open Prices

### Part of 
- #934 #935


src/components/ChallengeTakePicturesCard.vue
```diff
@@ -1,13 +1,23 @@
 <template>
-  <v-card v-if="exampleProofUrl">
+  <v-card v-if="challenge.exampleProofUrl">
     <v-card-title>
       {{ $t('Challenge.StepTakePictures.Title') }}
     </v-card-title>
     <v-card-text>
+      <v-row class="mb-2">
+        <v-col cols="12" sm="6">
+          <StatCard :value="challenge.numberOfProofs" variant="outlined" :subtitle="$t('Common.PicturesAdded')" />
```
raphodn, line 9: I'd keep the look & feel of stats as is (currently used many places elsewhere), and think of a new display separately ?

src/components/ChallengeTimeline.vue
```diff
@@ -19,7 +19,7 @@
   </v-timeline>
   <v-progress-linear
     style="width: 50%; margin-left: 25%; top: -55px; margin-top: -25px"
-    color="primary"
+    color="secondary"
```
raphodn, line 22: fyi the "secondary" color is used nowhere else in the app :sweat_smile: 
rather use "success" (green) here as a fallback

src/components/StatCard.vue
```diff
@@ -2,7 +2,7 @@
   <v-card
     :title="value"
     :subtitle="subtitle"
-    variant="tonal"
+    :variant="variant"
```
raphodn, line 5: linked to above comment, rather do this change in a separate PR ?

raphodn (approved):

src/components/ChallengeTimeline.vue
TTalex: Ok, done.

src/components/ChallengeTakePicturesCard.vue
TTalex: Sounds good, rolling back that change

PR #1339, opened 2025-01-26
feat(Stats): new Experiments section, with challenges & prices with price tag count

raphodn:
### What

see title & screenshot

### Screenshot

image

### DIFF
--- a/src/views/Stats.vue
+++ b/src/views/Stats.vue
@@ -89,6 +89,22 @@
     </v-col>
   </v-row>
 
+  <br>
+
+  <h2 class="text-h6 mb-1">
+    <v-icon size="x-small" icon="mdi-test-tube" />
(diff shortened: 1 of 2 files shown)


src/views/Stats.vue
```diff
@@ -89,6 +89,22 @@
...
+  <br>
+
+  <h2 class="text-h6 mb-1">
+    <v-icon size="x-small" icon="mdi-test-tube" />
+    {{ $t('Common.Experiments') }}
+  </h2>
+
+  <v-row>
+    <v-col cols="6" sm="4" md="3" lg="2">
+      <StatCard :value="1" :subtitle="$t('Common.Challenges')" />
```
raphodn, line 101: value hardcoded for now, until we start storing the Challenge metadata in a backend model

PR #1359, opened 2025-02-03
refactor(ContributionAssistant): Add support for new Not A Price option on canvas

TTalex:
### What
- Follow-up to https://github.com/openfoodfacts/open-prices-frontend/pull/1357
- Show this new option
- Also some tiny refactoring to use PRICE_TAG_STATUS_... constants

### Screenshot
Sans titre


src/i18n/locales/en.json
```diff
@@ -429,6 +429,7 @@
 			"PriceTagWithPrice": "Price tag with a price",
 			"PriceTagTruncated": "Price tag marked as truncated",
 			"PriceTagUnreadable": "Price tag marked as unreadable",
+			"PriceTagNotAPrice": "Price tag marked as not a price",
```
raphodn, line 432: wondering if we should repeat "Price tag" everywhere, or have a shorter version ?

src/components/ContributionAssistantPriceFormCard.vue
```diff
@@ -29,15 +29,15 @@
           </v-btn>
         </template>
         <v-list>
-          <v-list-item :slim="true" prepend-icon="mdi-eye-off-outline" @click="removePriceTag(2)">
+          <v-list-item :slim="true" prepend-icon="mdi-eye-off-outline" @click="removePriceTag(constants.PRICE_TAG_STATUS_UNREADABLE)">
```
raphodn, line 32: didn't know you could call constants in the template like this :eyes:

raphodn (approved):

src/components/ContributionAssistantPriceFormCard.vue
TTalex: Now that you mention it, it does seem a bit odd. Let's hope there are no side effects

src/i18n/locales/en.json
TTalex: I was wondering the same and almost removed it.
But I wasn't sure if "without a price" was understandable since there is no verb. ("Marked as unreadable" would be ok)

src/components/ContributionAssistantPriceFormCard.vue
raphodn: ok i think it doesn't work, reverting

PR #1368, opened 2025-02-08
feat(LocationMap): Display tooltips as cards and allow selection

TTalex:
### What
- Prettier tooltips as cards, with theme support
- Added an optional (hidden by default) select location button in tooltip, so users do not have to look in the list

### Screenshot
| Before | After |
| ------- | ----- |
| Screenshot 2025-02-08 at 17-07-06 Open Prices | Screenshot 2025-02-08 at 16-59-19 Open Prices |

### DIFF
--- a/src/components/LeafletMap.vue
+++ b/src/components/LeafletMap.vue
@@ -3,11 +3,31 @@
     <l-tile-layer :url="tiles" layer-type="base" name="OpenStreetMap" :attribution="attribution" />
     <l-marker v-for="location in locations" :key="getLocationUniqueID(location)" :lat-lng="getLocationLatLng(location)">
       <l-popup>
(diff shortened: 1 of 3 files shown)


src/components/LeafletMap.vue
```diff
@@ -3,11 +3,31 @@
...
+          <v-card-actions v-if="showActions">
+            <v-spacer v-if="$vuetify.display.smAndUp" />
+            <v-btn
+              class="float-right"
+              color="primary"
+              variant="flat"
+              :block="!$vuetify.display.smAndUp"
+              @click="locationSelected(location)"
+            >
+              {{ $t('Common.Select') }}
```
raphodn, line 27: I simplified to just "Select"

raphodn (approved):

PR #1382, opened 2025-02-16
feat(Barcode search): allow wildcard search thanks to search-a-licious

raphodn:
### What

Allow users to search for barcodes with wildcard.
"Hidden" feature, not advertised.

The product data format is slightly different, did some super quick hacks in the meantime.

### Screenshot

|Regular search|With wildcard|With 2 wildcards|
|-|-|-|
|image|image|image|


raphodn:
Actually this doesn't work (yet), I'm getting CORS errors. Thought it was only because of localhost, but it also doesn't work from https://prices.openfoodfacts.net

PR #1384, opened 2025-02-19
fix(Challenge): Timeline: fix off-by-one end date error

TTalex:
### What
- After mid-day, challenge timeline shows that the challenge ended, when there's still a day left.
- It was only a visual bug, with no effects on stats, as far as I can tell.
- This fixes it

### DIFF
--- a/src/components/ChallengeTimeline.vue
+++ b/src/components/ChallengeTimeline.vue
@@ -41,7 +41,9 @@ export default {
       return Math.round((new Date(this.challenge.endDate) - new Date(this.challenge.startDate)) / (1000 * 60 * 60 * 24))
     },
     todayIndex() {
-      return Math.round((new Date() - new Date(this.challenge.startDate)) / (1000 * 60 * 60 * 24))
+      let today = new Date()
+      today.setHours(0, 0, 0, 0)
+      return Math.round((today - new Date(this.challenge.startDate)) / (1000 * 60 * 60 * 24))
     },
     daysLeftText() {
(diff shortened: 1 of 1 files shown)


src/components/ChallengeTimeline.vue
```diff
@@ -41,13 +41,16 @@ export default {
...
-      return Math.round((new Date() - new Date(this.challenge.startDate)) / (1000 * 60 * 60 * 24))
+      let today = new Date()
+      today.setHours(0, 0, 0, 0)
+      return Math.round((today - new Date(this.challenge.startDate)) / (1000 * 60 * 60 * 24))
     },
     daysLeftText() {
       const daysLeft = this.nbDays - this.todayIndex
-      if (daysLeft <= 0) return "Challenge over"
+      if (daysLeft < 0) return "Challenge over"
+      if (daysLeft === 0) return "Last day"
```
raphodn, line 51: the API queries use `lte` so the last day should be counted as well :)

src/components/ChallengeTimeline.vue
```diff
@@ -41,13 +41,16 @@ export default {
...
       if (daysLeft >= this.nbDays) return "Not started"
-      return `${daysLeft} days left`
+      return `${daysLeft} day${daysLeft > 1 ? 's' : ''} left`
```
raphodn, line 53: might as well make it singular the day before last ^^
todo: translations instead

raphodn (approved):

PR #1394, opened 2025-02-23
feat(Experiments): Receipt Assistant

TTalex:
### What
- A new Experiment proposition to handle receipts
- Uses new receipt AI predictions added in PR https://github.com/openfoodfacts/open-prices/pull/715
 

### Screenshot
Screenshot 2025-02-23 at 19-53-33 Open Prices

### Video

https://github.com/user-attachments/assets/77d9898a-8e5a-4219-8be7-14d2374bffee

### Notes
- The table bypasses some of the usual price addition tools, allowing users to directly input price and barcode.
- The action column still allows for more complex edition
- Only items with both price and product are considered for addition
- Barcode/category field is smart
   - when empty, it offers a Scan icon to quickly open scan dialog (on mobile)
   - when filled, it offers a Search icon to quickly find a product matching barcode (enter key also works)
   - when a product is found, it is displayed instead of the field
- Some context is also displayed under products names.
   - Mainly, if a price with the same product_name was already added to the proof, a warning message is displayed. And a specific button has to be pressed at the bottom to add these possible duplicates.
- Like the Contribution Assistant `?proof_ids=50` query param can be used to skip the proof selection step
- Most likely full of bugs :D


teolemon:
1. Wow, just wow :-)
2. Being able to download a copy of the receipt would be nice. 
3. Being able to store quantities would be nice. 
4. Being able to store the total number of items on the receipt for consistency check would be nice
5. Being able to scan all the barcodes in bulk, fetch the product names using the API and attempt to auto-associate would be batshit crazy. 
6. What does it look like on a mobile phone ?

TTalex:
> 6. What does it look like on a mobile phone ?

It displays rows one after another, it's pretty long with a lot of items.
https://github.com/user-attachments/assets/1e8d0fc0-3cfc-4884-9b6d-17592d9c1a1a

teolemon:
@TTalex a little filter system perhaps ?

TTalex:
Update: this PR now supports the new ReceiptItem model introduced in https://github.com/openfoodfacts/open-prices/pull/778

TTalex:
Feedback:

- Remove datatable footer
- Move Add a new price button to the bottom, rename it "Add a missing price" or "Add a missing item"
- Hide one of the validation button when there are no existing prices, since they are redundant
- Force users to make a choice on product type, to avoid erroneous adds with detected categories.
    - We could force everything on barcodes, and prefil categories in the popup
- There seem to be some sync issues with the edit popup
- Edit popup UI: add a header and a close "X" in the top-right
- Rename barcode/category column to "product", with "Barcode or Category" as a subtitle
- Swap price and product columns (product should be first)
- V2? Add a confirmation button on delete, since buttons are pretty close to each other (think mobile)
- V2? Adding a new item could make use of the edit popup

raphodn:
Thanks for listing the feedbacks here ! No need to do all of them for the v1 :) just the couple bugs/buttons removal. And we can iterate together in futur PRs with other feedbacks 🙏

raphodn (approved):

PR #1435, opened 2025-03-08
feat(Proofs): display the new owner_consumption field

raphodn:
### What

Thanks to the addition of the `Proof.owner_consumption` field in the backend - https://github.com/openfoodfacts/open-prices/issues/583 - we add it in the Proof Upload card/form.

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -354,6 +354,7 @@
 		"Receipts": "Receipts",
 		"ReceiptPriceCount": "Number of prices",
 		"ReceiptPriceTotal": "Total amount",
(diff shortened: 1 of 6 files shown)


src/i18n/locales/en.json
```diff
@@ -354,6 +354,7 @@
 		"Receipts": "Receipts",
 		"ReceiptPriceCount": "Number of prices",
 		"ReceiptPriceTotal": "Total amount",
+		"ReceiptOwnerConsumption": "Count into my consumption stats",
```
raphodn, line 357: will need help on the wording ^^

raphodn: alternative: "Track in my consumption dashboard"

PR #1446, opened 2025-03-11
feat(Proofs): display the new owner_comment field (form, chip, history)

raphodn:
### What

Thanks to the addition of the `Proof.owner_comment` field in the backend - https://github.com/openfoodfacts/open-prices/pull/748 - we display it in the frontend
- Proof upload form (link + toggle)
- Proof footer chip (simple chip, similar to #1435)
- Proof history card

### Screenshot

|Page|Image|
|-|-|
|Proof form (create)|image|
|Proof form (after toggle)|image|
|Proof footer|image|
|Proof history|image|

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -159,6 +159,7 @@
 	},
 	"Common": {
 		"About": "About",
+		"AddComment": "Add a comment",
 		"AddPrice": "Add a price",
 		"AddPrices": "Add prices",
 		"AddNewPrice": "Add a new price",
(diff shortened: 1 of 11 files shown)


src/i18n/locales/en.json
```diff
@@ -198,6 +199,7 @@
 		"Chart": "Chart",
 		"Clear": "Clear",
 		"Contributors": "Contributors",
+		"Comment": "Comment",
```
raphodn, line 202: alternative: `Comment (public)` to remind that everything is public in OP ^^

PR #1459, opened 2025-03-21
feat(Challenges): read from api, add challenges list page

TTalex:
### What
- Following backend support for challenges https://github.com/openfoodfacts/open-prices/issues/691
- Frontend now reads the results from the API instead of a json
- Challenge page can be loaded without a param `/challenge` to load the first challenge with status ONGOING.
- Challenge page can be loaded with a param `/challenge/1` to load any challenge in db
- A page was added to list all challenges, with an option to show challenged with the ONGOING status (disabled by default)
- This is the page used in the side menu
- The Home banner, if displayed, still points to the current challenge

### Screenshot
The challenge page remains the same as before.

New challenges list page:
Capturechallenges


src/views/ChallengeList.vue
```diff
@@ -0,0 +1,132 @@
...
+        {{ $t('Challenge.ChallengeCount', { count: challengeTotal }) }}
+      </v-chip>
+      <LoadedCountChip :loadedCount="challengeList.length" :totalCount="challengeTotal" />
+      <FilterMenu kind="challenge" :currentFilter="currentFilter" :currentType="currentType" @update:currentFilter="toggleChallengeFilter($event)" @update:currentType="toggleChallengeType($event)" />
+    </v-col>
+  </v-row>
+
+  <v-row class="mt-0">
+    <v-col v-for="challenge in challengeList" :key="challenge" cols="12" sm="6" md="4" xl="3">
+      <v-card :id="'challenge_' + challenge.id" :class="'border-transparent'" @click="goToChallenge(challenge)">
```
raphodn, line 14: could be moved to a dedicated component (future PR)

src/views/ChallengeDetail.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <v-row>
+  <v-row v-if="!challenge && !loading">
```
raphodn, line 2: having breadcrumbs would be useful

src/views/ChallengeList.vue
```diff
@@ -0,0 +1,132 @@
...
+      loading: false,
+      currentFilter: '',
+      currentType: '',
+      currentOrder: 'id',
+    }
+  },
+  computed: {
+    getChallengesParams() {
+      let defaultParams = { order_by: this.currentOrder, page: this.challengePage, status: "ONGOING" }
+      if (this.currentFilter === 'show_all_status') {
```
raphodn, line 69: not sure if having this filter (and especially filtered by default) is a good idea. rather load them all but display them in sections ? I was surprised on first load to have 0 challenges being displayed

src/i18n/locales/en.json
```diff
@@ -155,7 +155,16 @@
...
+		"MostRecentContributions": "Most recent contributions",
+		"ChallengeCount": "{count} challenges | {count} challenge | {count} challenges",
+		"ChallengeStatus": "Status: {status}",
+		"ChallengeStatuses": {
+			"DRAFT": "Draft",
+			"UPCOMING": "Upcoming",
+			"ONGOING": "Ongoing",
+			"COMPLETED": "Completed"
+		},
+		"NoChallengeCurrentlyOngoing": "No challenge currently ongoing, please come back later !"
```
raphodn, line 167:
```suggestion
		"NoChallengeCurrentlyOngoing": "No challenge currently ongoing, please come back later!"
```

src/views/ChallengeList.vue
```diff
@@ -0,0 +1,132 @@
...
+    <v-col v-for="challenge in challengeList" :key="challenge" cols="12" sm="6" md="4" xl="3">
+      <v-card :id="'challenge_' + challenge.id" :class="'border-transparent'" @click="goToChallenge(challenge)">
+        <template #title>
+          {{ challenge.icon }} {{ challenge.title }} {{ challenge.subtitle }}
+        </template>
+
+        <v-divider />
+
+        <v-card-text>
+          <p class="mb-2">
```
raphodn, line 22: i think a chip with just the status info is enough

src/views/ChallengeList.vue
raphodn: also, mouse hover doesn't show it's a url (but this is recurring problem due to the usage of @click instead of <a> or router-link, we'll need to find a better way for all of these occurences)

raphodn (approved):

PR #1466, opened 2025-03-23
refactor(Challenges): Remove status filter + cleanup

TTalex:
### What
- Following suggestion in https://github.com/openfoodfacts/open-prices-frontend/pull/1459#discussion_r2008479596
- Drop the unnecessary filter, we won't have many challenges after all, might as well display them all

### Screenshot
Capturechallenges


raphodn (approved):
cool ! I also added a quick "cleanup" commit :)

PR #1501, opened 2025-04-11
refactor(Receipt Assistant): show a readonly toggle indicating that AI will run on the image

raphodn:
### What

Following #1394, we want to be more explicit to users about the use of AI.

We'll need to manage this in the backend.

For now we just show a readonly message.

### Screenshot

image

### DIFF
--- a/src/components/ProofMetadataInputRow.vue
+++ b/src/components/ProofMetadataInputRow.vue
@@ -105,6 +105,30 @@
       />
     </v-col>
   </v-row>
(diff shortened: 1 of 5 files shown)


TTalex (approved):
Looks good !

Got a bit confused by the disabled button, since the grey color made me think it was not selected (allow = false).

We could also add the `:assistedByAI="true"` to the ContributionAssistant when receipt is selected there.

raphodn:
@TTalex replaced the toggle with an "info alert" in #1506

PR #1513, opened 2025-04-17
feat(Challenge): Add support for multiple product categories

TTalex:
### What
- Challenge infos are loaded from the backend
- The `categories` field can contain multiple, comma-separated categories
- Until now, only the first category was used when displaying challenge stats and latest products
- This PR simply loops over the categories and calls the API multiple times

### Note
- Another option would be to query multiple categories at once, see https://github.com/openfoodfacts/open-prices/issues/760

### DIFF
--- a/src/views/ChallengeDetail.vue
+++ b/src/views/ChallengeDetail.vue
@@ -98,35 +98,36 @@ export default {
         }
       })
     },
-    getStats() {
+    async getStats() {
       this.loading = true
-      api.getPriceStats({ ...this.defaultParams, product__categories_tags__contains: this.challenge.categories[0] })
-      .then((data) => {
-        this.challenge.numberOfContributions = data.price__count
-        this.loading = false
-      })
+      let priceCount = 0
+      for (let i = 0; i < this.challenge.categories.length; i++) {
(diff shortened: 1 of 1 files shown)


src/views/ChallengeDetail.vue
```diff
@@ -98,35 +98,37 @@ export default {
...
-        api.getPriceStats({ ...this.defaultParams, product__categories_tags__contains: this.challenge.categories[0], owner: this.username })
-        .then((data) => {
-          this.challenge.userContributions = data.price__count
-        })
-        api.getProofs({ ...this.defaultParams, owner: this.username, size: 1 })
-        .then((data) => {
-          this.challenge.userProofContributions = data.total
-        })
+        let userPriceCount = 0
+        for (let i = 0; i < this.challenge.categories.length; i++) {
```
raphodn, line 115: ok I'd rather have something cleaner ^^ (without the loop, without the async)
I'm going to have a look at the backend and unlock the multiple-category filtering :)

TTalex:
Thanks to the new overlap param in https://github.com/openfoodfacts/open-prices/pull/799, this PR no longer needs looping.

raphodn (approved):

PR #1527, opened 2025-04-30
refactor(Challenge): use new tags api param

TTalex:
### What
- Using https://github.com/openfoodfacts/open-prices/pull/821 and https://github.com/openfoodfacts/open-prices/pull/813
- We can now simplify API calls to fetch Proofs and Prices matching a challenge
- This also fixes the issue of the proof count tallying every proof, instead of ones matching the challenge

### DIFF
--- a/src/views/ChallengeDetail.vue
+++ b/src/views/ChallengeDetail.vue
@@ -72,11 +72,12 @@ export default {
       return utils.dateEndOfDay(this.challenge.end_date)
     },
     defaultParams() {
-      return { created__gte: this.startDateMidnight, created__lte: this.endDateMidnight }
+      return { 
+        created__gte: this.startDateMidnight, 
+        created__lte: this.endDateMidnight, 
+        tags__contains: `challenge-${this.challenge.id}`
+      }
     },
-    challengeCategoriesParam() {
-      return this.challenge.categories.join(',')
-    }
   },
   mounted() {
     this.getChallenge()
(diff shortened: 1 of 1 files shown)


raphodn (approved):
Cool !

src/views/ChallengeDetail.vue
```diff
@@ -72,11 +72,12 @@ export default {
       return utils.dateEndOfDay(this.challenge.end_date)
     },
     defaultParams() {
-      return { created__gte: this.startDateMidnight, created__lte: this.endDateMidnight }
+      return { 
+        created__gte: this.startDateMidnight,
```
raphodn, line 76: actually we don't need the created filters anymore :)

raphodn: aaah well it's an interesting question actually. in the backend the rule for "does the proof belong to the challenge" is "at least 1 of its prices belongs and was added during the challenge". but the proof could be older or newer to the challenge ^^

raphodn: i'd rather set 1 single rule, in the backend, and have the frontend fetch only via tags.
and we can make the rule evolve in the backend of course

TTalex: Agreed, all rules should be in backend, I'll fix this in a future PR

raphodn: done :grin: https://github.com/openfoodfacts/open-prices-frontend/commit/08217745a3c779e0a2a2c551a035eca010ff374c

PR #1542, opened 2025-05-02
refactor(Receipt Assistant): show price count & price total below table

raphodn:
### What

In the new Receipt Assistant (following #1501), we add the `ProofReceiptPriceCountChip` & `ProofReceiptPriceTotalChip` components at the footer of the price table.

### Screenshot

|Before|After|
|-|-|
|image|image|

### DIFF
--- a/src/components/ReceiptTableCard.vue
+++ b/src/components/ReceiptTableCard.vue
@@ -49,6 +49,10 @@
 
     <v-card-actions>
       <v-row>
+        <v-col cols="6">
+          <ProofReceiptPriceCountChip v-if="proof && items" class="mr-1" :uploadedCount="items.length" :totalCount="proof.receipt_price_count" />
+          <ProofReceiptPriceTotalChip v-if="proof && items" :uploadedCount="proofPriceListSum" :totalCount="proof.receipt_price_total" :currency="proof.currency" />
+        </v-col>
         <v-spacer />
         <v-col>
           <v-btn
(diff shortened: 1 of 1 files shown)


src/components/ReceiptTableCard.vue
```diff
@@ -135,9 +141,13 @@ export default {
       ],
     }
   },
-
   computed: {
     ...mapStores(useAppStore),
+    proofPriceListSum() {
+      return this.items.reduce((acc, price) => {
+        return acc + parseFloat(price.predicted_data.price)
```
raphodn, line 148: @TTalex do you think we should filter only on valid products ?
for instance in my current code/screenshot the count & sum are done on the ALL items in the table

raphodn: I'm asking because I felt a bit of confusion. but I think it's because I'm used to see these stats in the `PriceAlreadyUploadedListCard` component, whereas here the prices are not necessarily uploaded yet.. 

On the other hand it's super useful to see if the AI has correctly predicted the count & sum of the receipt !

raphodn: On the other hand it's super useful to see if the AI has correctly predicted the count & sum of the receipt !

TTalex: Upload button only counts valid prices, so that would make sense to do the same here.
Or maybe use the fact that we have two numbers, with the valid ones on the left and the total on the right ? Don't know if it causes more confusion than anything.

raphodn: merging as is, we'll see how it feels and improve on top of it

PR #1549, opened 2025-05-06
fix(Products): better manage non-en category tags

### LINKED ISSUE #1544: Products page: add support for categories not starting with "en"
### Problem
* Some categories are language specific, with no english equivalent.
* In those cases, `id`s in data/categories/xx.json do not start with `en:`
* Users are allowed to create prices with these categories, example: https://prices.openfoodfacts.org/prices/102750
* Yet, the frontend fails to display them: https://prices.openfoodfacts.org/products/fr:batavia
* This is due to checks on product, and only looking at categories if it starts with "en"

### Proposed solution
* We could probably replace the check with the presence of ":"
* This would need to change in the following files
https://github.com/openfoodfacts/open-prices-frontend/blob/aebf7926ffc9b1797c5acc1b3912d5500893a8e4/src/components/ProductInputRow.vue#L195
https://github.com/openfoodfacts/open-prices-frontend/blob/aebf7926ffc9b1797c5acc1b3912d5500893a8e4/src/views/ProductDetail.vue#L118

raphodn:
### What

Fixes #1544 (allowing to display non-en category tags, such as https://prices.openfoodfacts.org/products/fr:batavia)

### DIFF
--- a/src/components/ProductInputRow.vue
+++ b/src/components/ProductInputRow.vue
@@ -190,7 +190,7 @@ export default {
   },
   mounted() {
     if (this.$route.query.code) {
-      if (this.$route.query.code.startsWith('en')) {
+      if (this.$route.query.code.includes(':')) {
         this.productForm.type = constants.PRICE_TYPE_CATEGORY
         this.productForm.category_tag = this.$route.query.code
       }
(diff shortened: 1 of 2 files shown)


TTalex:
@raphodn Nice, note that it still shows "category not found"
Screenshot from 2025-05-07 18-58-00

PR #1601, opened 2025-07-22
feat(Proof Upload): add an action to 'Add prices' if the user does not want to use the community validation

raphodn:


src/views/ProofPriceTagAddMultiple.vue
```diff
@@ -74,19 +83,27 @@ export default {
...
+      firstProofUploaded: null,
       proofUploadCount: 0
     }
   },
   computed: {
     getPriceValidationUrl() {
       return '/experiments/price-validation-assistant'
     },
+    getPriceAddMultipleProofIdUrl() {
+      return `/prices/add/multiple?proof_ids=${this.firstProofUploaded.id}`
```
raphodn, line 95: need to do another PR to make this work in `PriceAddMultiple`

raphodn: actually the logic was already implemented. but i moved to it the correct view in #1602

Issue #1587, opened 2025-07-07
Improve visibility and placement of "Upload multiple price tags" [UX]


julia-gx:
### Problem

**Flow [web]: Contribute --> From a price tag --> Upload multiple price tags** 

Currently, the option to upload multiple price tags (same shop and date) is placed in a blue banner below the main heading of the "upload proof" form, but not at a decision point in the flow. It’s presented too late. It takes the user to a different mode but its appearance, wording, and placement do not clearly indicate that. Looks like a note/tip, not a mode switch or primary action. 

As a result:
- users may not notice this alternative upload flow
- it looks like a passive info tip, not a navigational choice
- It appears almost mid-form, users if click it can suddenly navigate away, potentially losing progress of adding single price or getting confused about where they are.
- there's no indication that it leads to a new page or a mode change

Since it's an important optional path (for users uploading more than one price tag), its placement and styling should make sense in context without being confusing. 

### Proposed solution

Above the form, adding it as a toggle choice at the top, and present users with:

🔘 Upload single price tag  
⚪ Upload multiple price tags from same shop & date

If the user selects "multiple", a confirmation modal could warn:
“You will be redirected to the multi-upload page/mode. Any unsaved changes will be lost.”
[Continue] [Cancel]

### Additional context
Why it is an UX issue? 
- hidden navigation to a different mode is **not discoverable** or **predictable**
- this is a **primary decision** (single vs multiple upload) and should be treated as such
- mid-form branching causes **flow disruption** and **risk of data loss**

This change would improve: task discoverability, user confidence, flow clarity, data preservation

julia-gx:
You mean, having one “multiple proofs” mode that lets upload one or many proofs? So no toggle or separate paths? But in the single proof mode, there’s no AI involved, but in the multiple proof mode yes, so how to handle this?

raphodn:
I'm going to try to do a graph to visualize the current different possible steps

julia-gx:
**Step 1: What are you uploading?** 

- [ ]  Price Tag(s) 

- [ ]  Receipt(s) 

**Step 2: How do you want to enter price(s)?**

- [ ]  Let AI extract  

- [ ]  I’ll enter manually 

**Step 3: 
Users move to the data entry step where they complete or confirm extracted data.** 

For example: 
CASE 1: single price tag

_AI Extraction_

AI detects price and (optionally) product name from picture

User sees:
Extracted price (editable)
Dropdown: Currency
[...]

PR #1520, opened 2025-04-23
feat: new Contribute page

raphodn:
### What

WIP

PR #1641, opened 2025-08-17
feat(Proof Upload): make the PriceTagMultiple uploader the default

raphodn:
### What

- In #1520 we oriented users wanting to upload a price tag proof towards the "Price addition" workflow.
- In this PR we replace it with the "Proof addition" workflow
  - linked with #1598
- Linked to the discussion in #1587.

### Screenshot

|Page|Before|After|
|-|-|-|
|Contribute|image|same|
|"Price tag" selected|image|image|


teolemon (commented):
I'm sorry, but I fail to see how the Proof wording is more intuitive vs the Price Tag and Receipt thing.

PR #1650, opened 2025-08-30
feat(ReceiptAssitant): suggest product codes based on product names

TTalex:
### What
- Uses predicted generated with backend PR: https://github.com/openfoodfacts/open-prices/pull/952

### Screenshot
image

### Demo gif
Timeline-12(1)

### Notes
The current UI requires users to click on the suggested barcode to see associated product.
If the user is unhappy with the suggestion, they would have to edit the product using the edit button on the side, which might not be intuitive.

In the backend PR @teolemon suggested some other interaction, using hover or showing more info, not sure how to make it look nice.

### DIFF
--- a/src/components/ReceiptTableCard.vue
+++ b/src/components/ReceiptTableCard.vue
@@ -37,7 +37,7 @@
               <span
                 class="text-primary font-weight-medium"
                 style="cursor: pointer;"
-                @click="item.product_code = item.predicted_product_code"
+                @click="handleClickProductCodeSuggestion(item)"
               >
                 {{ item.predicted_product_code }}
               </span>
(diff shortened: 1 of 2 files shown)


src/components/ReceiptTableCard.vue
```diff
@@ -22,16 +22,26 @@
...
+              density="compact"
+              :rules="rules"
+              :append-inner-icon="item.product_code ? 'mdi-magnify' : 'mdi-barcode-scan'"
+              @click:append-inner="item.product_code ? findProduct(item) : launchBarcodeScanner(item)"
+              @keydown.enter="findProduct(item)"
+            />
+            <div v-if="item.predicted_product_code" class="text-caption">
+              {{ $t('Common.SuggestedBarcode') }}
+              <span
+                class="fake-link"
```
raphodn, line 38: changed the css to transform to link

image

PR #1704, opened 2025-09-27
fix(ProofPriceTagAssistant): Properly wait when new boxes are drawn on old proofs

TTalex:
### What
- ProofPriceTagAssistant has a feature skipping wait time when the proof is older than a day.
- This was introduced in PR https://github.com/openfoodfacts/open-prices-frontend/pull/1331 with the idea that it's pointless waiting for an AI process that probably failed
- However, this skip also happened when new boxes were drawn, leading to errors because the frontend did not wait long enough.
- This PR rectifies this, by only skipping the wait time when the proof is first loaded. And fully waiting when new boxes are drawn.

### DIFF
--- a/src/views/ProofPriceTagAssistant.vue
+++ b/src/views/ProofPriceTagAssistant.vue
@@ -406,16 +406,9 @@ export default {
         this.priceTags = []
         this.boundingBoxesFromServer = []
       } else {
-        let maxTries = 10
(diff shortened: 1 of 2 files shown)


raphodn (approved):

raphodn:
Issue created in my previous refactoring... https://github.com/openfoodfacts/open-prices-frontend/pull/1695

PR #1723, opened 2025-10-04
feat(PriceTags): replace barcode with similar one if invalid

TTalex:
### What
- When loading price tags predictions, use the best (first) similar barcode if available, and if barcode is invalid

Note: this could also be done in the backend ?

### DIFF
--- a/src/utils/proof.js
+++ b/src/utils/proof.js
@@ -85,6 +85,15 @@ function handlePriceTag(priceTag) {
     productPriceForm.price_is_discounted = selectedPrice ? selectedPrice.price_is_discounted : false
     productPriceForm.price_without_discount = selectedPrice.price_without_discount ? selectedPrice.price_without_discount.toString() : ""
     productPriceForm.discount_type = selectedPrice.discount_type || ""
+
+    // Check for good similar barcodes
+    if (label.similar_barcodes && label.similar_barcodes.length) {
+      productPriceForm.similar_barcodes = label.similar_barcodes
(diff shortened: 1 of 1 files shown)


src/utils/proof.js
```diff
@@ -85,6 +85,15 @@ function handlePriceTag(priceTag) {
     productPriceForm.price_is_discounted = selectedPrice ? selectedPrice.price_is_discounted : false
     productPriceForm.price_without_discount = selectedPrice.price_without_discount ? selectedPrice.price_without_discount.toString() : ""
     productPriceForm.discount_type = selectedPrice.discount_type || ""
+
+    // Check for good similar barcodes
+    if (label.similar_barcodes && label.similar_barcodes.length) {
+      productPriceForm.similar_barcodes = label.similar_barcodes
+      // if there is a similar barcode, it's probably better than the guessed one if it's invalid
+      if (!barcode_utils.isBarcodeValid(productPriceForm.product_code)) {
+        productPriceForm.product_code = label.similar_barcodes[0].barcode
```
raphodn, line 94: what about giving the user the choice ? for instance:
- replacing the "invalid" product section with a message/banner
- having the "Find the product" display the whole (or a subset) of the similar_barcodes list ?

raphodn: ah kinda like #1724 yeah ^^
+ we "force" the user to go through the BarcodeScannerDialog, instead of validating a vary-likely wrong product

TTalex:
Will not merge, see https://github.com/openfoodfacts/open-prices-frontend/pull/1724 for a better approach.

PR #1724, opened 2025-10-04
feat(BarcodeScannerDialog): Display similar barcodes as choices

TTalex:
### What

Following changes in the backend: https://github.com/openfoodfacts/open-prices/issues/1035

- Related to https://github.com/openfoodfacts/open-prices-frontend/pull/1723, could combine both PRs
- This PR displays all similar barcode products in the product selection dialog
- User can still search for the product using the barcode form, as before, but they now can select a product in the list

Todos:
- Maybe display only the 4 most similar products, or limit to very close similarity ? This would avoid busy UI and reduce API calls.
- Product display as a row could use some UI help

### Screenshot
Capture


TTalex:
TODO: Display clearly that these are suggested

TTalex:
@raphodn If you want to take a look at this one ;)

src/components/BarcodeScannerDialog.vue
```diff
@@ -52,6 +52,21 @@
               </v-text-field>
             </v-form>
             <ProductCard v-for="product in productSearchResultList" :key="product" :product="product" :hideCategoriesAndLabels="true" :hideActionMenuButton="true" :readonly="true" elevation="1" @click="barcodeSend(product.code)" />
+
+            <div v-if="barcodeManualInputSimilarBarcodes.length">
```
raphodn: tested, how it looks like 👍 

Image

src/components/BarcodeScannerDialog.vue
```diff
@@ -124,6 +143,7 @@ export default {
       },
       barcodeManualFormValid: false,
       productSearchResultList: [],
+      productSuggestionResultDict: {},
```
raphodn: why a dict instead of a list ?
(list are easier to loop on and "wrap around in your head" 😅 )

raphodn (approved):

src/components/BarcodeScannerDialog.vue
TTalex: The API and utils/proofs.js gives us a list `barcodeManualInputSimilarBarcodes` with ordered barcodes, from closest to farthest barcodes. The goal is to keep this list order when displaying it to users.

For each element of this list, we fetch the corresponding product using API. This is asynchronous, so we get results in random order. We could fill a list here, but it will break order.

So we end up using the original `barcodeManualInputSimilarBarcodes` as the looping list to display products. For each barcode, we need to find the product we just retrieved, that's where a dict is more efficient than a list.

---

Another option would be to save product info back into the original `barcodeManualInputSimilarBarcodes` list, but that's not clean either.
We could also use a list instead of a dict, and sort it before display ? Maybe it would be easier to understand ?

raphodn: looking into improving this 👍

raphodn: ok done! used a list, and kept order : https://github.com/openfoodfacts/open-prices-frontend/pull/1724/changes/b44cb8feb10d7f8bf2d30bb6fe7b344b118ebc2f

PR #1771, opened 2025-10-24
refactor(CreateProduct): Always display list of suggested missing products

TTalex:
### What
- Before this PR, the list of suggested missing products didn't show if user landed on experiment with a product_code as query param
- We now load the list every time
- PR also changes the way this first step is displayed, using both columns

### Screenshot
image

### Related issue
- https://github.com/openfoodfacts/open-prices/issues/1049

### DIFF
--- a/src/views/CreateOffProduct.vue
+++ b/src/views/CreateOffProduct.vue
@@ -57,7 +57,7 @@
     <v-col cols="12" md="6">
       <v-card
         class="mb-4"
-        :title="$t('CreateOffProduct.ChooseUnknownProductGuide')"
+        :title="$t('CreateOffProduct.SelectUnknownProductGuide')"
         prepend-icon="mdi-tag-plus-outline"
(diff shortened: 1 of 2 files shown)


src/views/CreateOffProduct.vue
```diff
@@ -355,7 +368,7 @@ export default {
     stepItemList() {
       return [
         {
-          title: this.$t('Common.BarcodeType'),
+          title: this.$t('CreateOffProduct.SelectUnknownProduct'),
```
raphodn: In this commit - https://github.com/openfoodfacts/open-prices-frontend/pull/1771/commits/2d843352bb2673797d2d78be53e6d18de998b2de - I renamed "Choose" to "Select", as we use it pretty much everywhere else (1 exception, but thinking of changing as well to simplify)

raphodn (approved):

PR #1776, opened 2025-10-26
refactor(Proof upload): chain promises instead of compressing/uploading at the same time

raphodn:
### What

When multiple proofs are being uploaded, we weren't managing correctly the compression + upload combo.
Doing things pretty much in parallel, instead of synchronously.

This might also fix images not being uploaded in the selected order ? And image hash calculation to detect duplicates ??

### Ideas to improve

- show a text to explain what is happening
- separate compression & upload steps ?

### DIFF
--- a/src/components/ProofUploadCard.vue
+++ b/src/components/ProofUploadCard.vue
@@ -274,9 +274,9 @@ export default {
     uploadProofList() {
       this.step = 2
       // loop on images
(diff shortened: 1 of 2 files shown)


TTalex:
Doesn't Promise.all still run everything in parallel ?

raphodn:
> Doesn't Promise.all still run everything in parallel ?

yeah you're probably right... :facepalm:

PR #1825, opened 2025-11-13
feat(Moderation): allow logged-in users to flag proofs

raphodn:
### What

Similar to #1823

### Screenshot

||Flag modal|
|-|-|
||image|

### DIFF
--- a/src/components/ModerationFlagCreateDialog.vue
+++ b/src/components/ModerationFlagCreateDialog.vue
@@ -11,6 +11,7 @@
         <v-row>
           <v-col cols="12">
             <PriceCard v-if="price" :price="price" :product="price.product" :hidePriceFooterRow="false" :hideActionMenuButton="true" :readonly="true" />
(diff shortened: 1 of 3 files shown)


julia-gx:
hey, maybe making the “reason” label slightly more prominent (for example semibold, not bold), because this is a primary user action there and right now it doesn't stand out. 
Also, grouping related elements with tighter spacing. Currently, it seems visually “loose" a bit :)

raphodn:
> hey, maybe making the “reason” label slightly more prominent (for example semibold, not bold), because this is a primary user action there and right now it doesn't stand out. Also, grouping related elements with tighter spacing. Currently, it seems visually “loose" a bit :)

hey thanks for the feedback !
these modals are definitely to be improved:
- first i need to make a PR adding all the possible "flag" reasons
- and theses edit/delete/flag dialogs need to made sexyer, without the seperation :ok_hand:

PR #1827, opened 2025-11-13
feat(Proof Upload): if receipt, add an action step + default to the Receipt Assistant

raphodn:
### What

New `ProofReceiptAdd` component.
- launched from the Contribute page (when selecting "receipt")
- after the receipt is uploaded, "Add prices" action redirects to the Receipt Assistant

### Todo

Improve the "Action" page to make it fluid for users
- make the default action more visible (color): #1961
- add help text: #1910

### Screenshots

||Before|After|
|-|-|-|
|url|`/prices/add/multiple?proof_type=RECEIPT`|`/proofs/add/receipt`|
|image|image|step 1<br>image<br>step 2<br>image|

### DIFF
--- a/src/views/ProofPriceTagAddMultiple.vue
+++ b/src/views/ProofPriceTagAddMultiple.vue
@@ -50,7 +50,6 @@
         @click="reloadPage"
       />
     </v-col>
-    
     <v-col cols="12" sm="6" lg="4">
       <v-card
         :title="$t('Common.MyDashboard')"
--- a/src/views/ProofReceiptAdd.vue
+++ b/src/views/ProofReceiptAdd.vue
@@ -42,7 +42,6 @@
         @click="reloadPage"
       />
     </v-col>
-    
     <v-col cols="12" sm="6" lg="4">
       <v-card
         :title="$t('Common.MyDashboard')"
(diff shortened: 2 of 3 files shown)


TTalex:
The idea of the new component is to split the process in two ? Allow users to stop after proof upload ?

When I first looked at it, I thought it was to have a centralized receipt upload form, and let users decide to use the assistant or not after upload. But currently it looks like it just replaces the first assistant step, with the only difference to disallow selecting a previous proof ?

I'm not against it, just wanted to understand the logic.

raphodn:
> The idea of the new component is to split the process in two ? Allow users to stop after proof upload ?
> 
> When I first looked at it, I thought it was to have a centralized receipt upload form, and let users decide to use the assistant or not after upload. But currently it looks like it just replaces the first assistant step, with the only difference to disallow selecting a previous proof ?
> 
> I'm not against it, just wanted to understand the logic.

Yes indeed, i switched the PR back to draft because I'm not completely satisfied yet..
- the action step will probably be cumbersome for existing users that are used to having a direct workflow..
- but in case there will always be a few seconds (minute ?) delay because the AI needs to extract data from the image
- so having this intermediate Actions page could be helpful to give more context
- it's in my opining also helpful technically, separating the upload step (1 or multiple price tags or receipts), and the addition step
- selecting a previous proof can be done via the card menu, no ?

PR #1685, opened 2025-09-21
feat(Experiment): Add product to product opener using prices information

TTalex:
### What
This is a proof of concept.
This experiment helps creating products in product opener by displaying a simple form, pre-filled with info it gathered from existing prices (product name, stores, countries).
It also displays the proofs containing the item, in a zoomable way, to find more info on the product.
Product front images can be cropped directly from the displayed proof.
Product categories are suggested using challenge categories.

If the product already exists, the experiments limits the number of available fields, but allows quick modifications.

This experiment requires a new library (vue-zoomable), as well as changes on backend (PR in progress)

TODOs:

- Code is awful, refactoring is needed if the POC is validated
- Text strings are temporary, proper translation values need to be added

### Screenshot
image

### Demo video (with audio commentary !)
https://github.com/user-attachments/assets/15bd0dd3-89b4-4170-a135-9f70bb8723a9

PR #1959, opened 2026-01-03
refactor(Product Detail): if product missing, new promo banner to talk about the CreateOffProduct experiment

### LINKED ISSUE #1336: A page to create missing products
### Story

Users regularly add prices for products that are not yet in the OxF database.
We usually have a few info regarding this product: name, picture, brand, weight (from the proof image, and the price tag predictions).

### How

- a frontend page that lists products to add
  - image = draw a bounding box from the proof (we'll need a cropping mecanism)
  - product_name = extracted from the PriceTag prediction
  - other fields: simple form
- the product form, when filled, is sent to the Open Prices backend, who then sends an authenticated request to the OFF server (like robotoff does it. how exactly ??)

### Related issues

- https://github.com/openfoodfacts/open-prices/issues/610
- #459
- #604
- We need to improve the sync first ?
  - https://github.com/openfoodfacts/open-prices/issues/551
  - Only OFF products are synced live with Redis (other flavors are synced daily)

raphodn:
### What

In #1685 we created a new experiment allowing users to create new products directly to Open Food Facts using Open Prices data.

Here we add a promo banner to this page `CreateOpenFoodFactsProductPromoBanner`
- on the product details page
- only if the product is not found
- and it has at least 1 price

### Screenshot

|Before|After|
|-|-|
|image|Screenshot from 2026-01-03 22-37-06|

### DIFF
--- a/src/components/CreateOffProductPromoBanner.vue
+++ b/src/components/CreateOffProductPromoBanner.vue
@@ -7,7 +7,7 @@
     @click="$router.push(url)"
   >
     <v-banner-text style="padding-inline-end:10px;">
-      {{ $t('ProofAdd.PromoReceiptAssistant') }}
+      {{ $t('ProofAdd.ProductMissingPromoBanner') }}
     </v-banner-text>
     <v-spacer /><!-- needed to push v-banner-actions to the right on big screens -->
     <v-banner-actions>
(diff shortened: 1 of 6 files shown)


teolemon (commented):
Nice, it would almost be tempting to remove the above banner.

Issue #1992, opened 2026-01-19
Location emoji not display correctly anymore


NancyWei123:
### What

Uncaught (in promise) TypeError: Cannot destructure property 'type' of 'vnode' as it is null.
### Steps to reproduce the behavior:
Open developer tools and click location of a product.
The page cannot redirect correctly and there are mistakes.

### Expected behavior
Display locations correctly.

### Screenshots
<img width="1908" height="953" alt="Image" src="https://github.com/user-attachments/assets/73b75943-cfc0-4abb-8d5e-f441d0bf71d3" />

### Platform (Desktop, Mobile, Hunger Games)
 Windows 10

TTalex:
Good catch !

Another user has also reported an issue that stems from the same source (see picture below)

<img width="331" height="719" alt="Image" src="https://github.com/user-attachments/assets/d14739fb-1051-43ab-8ec6-5278d3f61ab2" />

I linked a PR that should fix both issues at once https://github.com/openfoodfacts/open-prices-frontend/pull/1995

raphodn:
Thanks to you 2 ! 
Was due to #1989 
Fixed in https://github.com/openfoodfacts/open-prices-frontend/pull/1995

PR #1994, opened 2026-01-19
fix(null error): fix iuuse#1992

NancyWei123:
### What
- 
fix issue Uncaught (in promise) TypeError: Cannot destructure property 'type' of 'vnode' as it is null.
### Screenshot

before
image

after
image

### Fixes bug(s)
- 
#1992 
### Part of 
- 
#1992


NancyWei123:
Sure

Issue #1974, opened 2026-01-15
Receipt Assistant: allow users to add prices without AI


raphodn:
### Story

AI is the single & default path.
But for PriceTag proofs we offer the possibility to add prices without AI (or at least without the community validation, slightly different).

We could think of a toggle, and bypass the AI functionality. 
We should then also skip the AI in the backend as well ?

PR #2004, opened 2026-01-23
feat(): allow users to add prices without AI

NancyWei123:
### What
User can choose to don't use ai when adding prices.
### Screenshot
image
### Part of 
#1974

### DIFF
--- a/src/views/ProofPriceTagAssistant.vue
+++ b/src/views/ProofPriceTagAssistant.vue
@@ -18,7 +18,7 @@
   <template v-if="step === 1">
     <v-row>
       <v-col cols="12" md="6">
-        <ProofUploadCard :typePriceTagOnly="true" @proof="onProofUploaded($event)" />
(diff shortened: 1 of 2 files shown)


src/views/ProofPriceTagAssistant.vue
```diff
@@ -410,6 +411,9 @@ export default {
       // proof image
       this.imageSrc = proof_utils.getImageFullUrl(proof.file_path)
 
+      if(this.skipAIValue) {
```
raphodn: why would you change the proof to RECEIPT here ?
PRICE_TAG & RECEIPT are 2 different types of proof images
you can have a look here : https://prices.openfoodfacts.org/proofs

the proof type, and the way to add prices (manually or AI assisted), are 2 very different things

NancyWei123:
Thank you! I have understanded the difference between PRICE_TAG & RECEIPT.

PR #2028, opened 2026-02-06
feat(ProofDetail): Allow displaying price tag bounding boxes

### LINKED ISSUE #2018: Proof detail: allow displaying the price tag bounding boxes (with a toggle ?)
### Story

We already have UI to view/edit the proof's price tags bounding boxes.

https://prices.openfoodfacts.org/experiments/proof-price-tag-assistant?proof_ids=69965

But on the proof detail page it is not available.

https://prices.openfoodfacts.org/proofs/69965

### Extra ideas

Panoramax has a similar feature to toggle the display of "tags".

TTalex:
### What
- Added an option to display bounding boxes on the Proof detail page
    - Option is only shown when bounding boxes do exist on the proof, and they are only loaded in the proof detail page
- No interactions yet, just simply show boxes, as displayed in the proof price tag assistant
- Toggle is done using Proof card menu or configured in user settings. Hidden by default.
- When bounding boxes are displayed, the image is replaced by a canvas. It stays as an image otherwise. I was worried that using canvas all the time would lead to performance or accessibility issues.

### Screenshot
Menu to display bounding boxes 
image

Same menu to hide shown bounding boxes
image

New option in the settings page
image

### Fixes bug(s)
- Fixes #2018


src/views/ProofDetail.vue
```diff
@@ -140,6 +141,14 @@ export default {
           this.priceTotal = data.total
         })
     },
+    getPriceTagsBoundingBoxes() {
+      return openPricesApi.getPriceTags({proof_id: this.proofId, size: 100}).then(data => {
+        if (!data.items) return
```
raphodn: i think there is an issue here.

src/components/ProofCard.vue
```diff
@@ -14,24 +14,39 @@
     <v-divider v-if="!hideProofHeader" />
 
     <v-card-text>
-      <v-img v-if="proof.file_path" :src="getProofImageFullUrl" :height="imageHeight" />
+      <ContributionAssistantDrawCanvas
```
raphodn: i did a test with a Price Tag proof. proof detail > menu > toggle display of price tags

Image

src/components/ProofActionMenuButton.vue
```diff
@@ -107,11 +111,16 @@ export default {
...
+    showPriceTagsBoundingBoxes: {
+      type: Boolean,
+      default: false
+    },
     style: {
       type: String,
       default: 'position:absolute;bottom:6px;right:0;'
     }
   },
+  emits: ['update:showPriceTagsBoundingBoxes'],
```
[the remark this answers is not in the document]
TTalex: Yes, I found out v-model is a [shortcut that includes the @update on a component](https://vuejs.org/guide/components/v-model).

In this specific case, there is an `@update:showPriceTagsBoundingBoxes="handleShowPriceTagsBoundingBoxes"` in the parent `ProofFooterRow`, but it's not required in the grandparent `ProofCard`, since v-model i used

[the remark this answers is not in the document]
raphodn: ok ! i'm rather in favor of explicit code than magic :stuck_out_tongue:

src/views/ProofDetail.vue
TTalex: Good catch, thanks

PR #2035, opened 2026-02-13
feat(UX): new HelpDialog components to display help info + add info text on Validate page

raphodn:
### What

Allow setting a "Help" text in the router config
- if set, it will display an :information_source: icon in the breadcrumbs
- clicking on it will open a modal

### Screenshot

||Image|
|-|-|
|breadcrumbs|image|
|modal|image|

### DIFF
--- a/src/i18n/locales/en.json
+++ b/src/i18n/locales/en.json
@@ -941,7 +941,7 @@
 		},
 		"Validate": {
 			"Title": "Validate",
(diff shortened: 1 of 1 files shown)


julia-gx:
There is a lack of information about the consequences of validation :)
So after this: 

> For each price below, check that the price, barcode, or category are correct, fix them if needed and confirm.

I'd add something like this: 

> When you confirm a price, it’s added to the public database and shared with the community.

PR #2080, opened 2026-03-15
fix(ReceiptAssistant): display an error when discounts are >= full price

### LINKED ISSUE #840: Discounted Price and Full Price should be allowed to be the same
Epic: #

### Description
I tried to add that price for `3450970060424` however I suppose since the price is the same when the Create button is pressed, it looks like the process is completed but the price doesn't appear in the Last prices page, nor on https://prices.openfoodfacts.org/app/products/3450970060424 the product page.

The product was displayed in a different part of the location/supermarket than the usual aisle ( a large aisle with a lot of products stacked up around ) and we can see the mention Promotion on the price tag picture used as proof. If the last price ( about a month ago ) of the product is checked on https://prices.openfoodfacts.org/app/products/3450970060424 it's the same as today discounted price.

I can't tell if it's mistake on the supermarket's part or an attempt to mislead the consumer.

https://prices.openfoodfacts.org/app/proofs/9024

### LINKED ISSUE #2067: Receipt Assistant: display an error when discounts are the same as full prices
### What
* API disallows prices created with same value between discounts and full price
* However, AI detection sometimes does put the same values
* This leads to price creation returning an (unhandled) error, and failing to create prices, without displaying any issue to the user.

Two things should be done:
1. After calling the API, handle price creation errors, and display to users when only part of the prices were created, and why
2. Before calling the API, consider discounted prices with same values as invalid, and show why. This should be handled the same way as a price missing, for example.

We could also automatically remove the discount flag and discounted price, but I'm not sure that it's clean. Usually the AI is right that the price was discounted, but failed to guess the value. So user input is required.

Example proof https://prices.openfoodfacts.org/experiments/receipt-assistant?proof_ids=92592

DevDs1989:
### What
Added a warning for when the discounted price is the same as full price thus prompting the user to take a look at it. Also added a check that validates if the two prices are same or not. To account for the new error a new key was added in the `i18n` en json file.

### Screenshot
image

### Fixes bug(s)
Fixes #2067 
Fixes #840


TTalex (approved):
Looks good, thanks ! Made tiny readability changes.

I'm also noticing that `:rules="priceRules"` is set on the price text field, but doesn't seem to be used (since we overwrite the class), something to fix another time.

Issue #230, opened 2024-01-26
Check at startup that the session cookie is valid


raphael0202:
Check using `GET /api/v1/session` that the cookie session is valid.
Otherwise, as one of the only part of the code requiring auth is the proof and price creation, we only know that the user has invalid auth cookie at this step, which is problematic.

Also, when receiving an HTTP Forbidden, logout and redirect to the login page.

raphodn:
I'm a bit lost between the cookies and the vue.js store. 

Currently the user token is stored in the store/localStorage. But with the cookie it seems that there is 2 different authentication mechanisms ? I'm facing a bug, I logged out and logged back in, but still have authentication problems, most likely because I am still logged in in OFF ??

raphael0202:
> Currently the user token is stored in the store/localStorage. But with the cookie it seems that there is 2 different authentication mechanisms ?

We first check the existence of a session cookie, then for the auth token.

The bug you're facing is when you try to implement cookie auth?

raphodn:
I had a look
- our store.js persistes data in the Local Storage. navigating between localhost, prices.openfoodfacts.net & prices.openfoodfacts.org does not provoke any conflict
- our frontend authenticates with the backend ONLY via header Bearer token
- but "with same-origin fetch, browser cookies are sent by default"
- so inspecting the frontend API calls, i see both header & cookies passed
- and sometimes the server answers saying the cookies are invalid, as we check for cookies first before header : https://github.com/openfoodfacts/open-prices/blob/main/open_prices/common/authentication.py

raphodn:
Better manage token invalid errors
- [x] stop sending domain cookies to the backend. Token header auth only: #2148
- [ ] if token error, redirect to login

raphodn:
I opened a PR - #2148 - to avoid sending cookies in the API calls, as the frontend app only uses header token as authentication.
But I'm wondering if it's something that could be fixed (instead ? or in parallel ?) in the backend
- if both cookie & token provided, and cookie auth fails, then check token as well instead of early fail ?

TTalex:
_(Copying my slack message, for reference)_
I think I've got it. Conflict isn't between OFF and Prices. It's between Django Admin and Prices.

Looking at our codebase, there was no way to actually store opsession cookie from regular login. Since we never set ?set_cookie=1 in auth query. (It's even worse, in authenticate.py file, we read it as a GET parameter .. but the query is a POST. So it's never true).

PR #2148, opened 2026-04-16
refactor(Auth): API: stop sending domain cookies to the API, they conflict with the token authentication

raphodn:
### What

The frontend's authentication mecanism with Open Prices API should be ONLY via token.
But with other OFF domains, we sometimes have cookies persisted.
And when sent to the Open Prices API, they are checked first instead of the token...

### Solution

Only interact with the Open Prices API using token. Remove cookies from the calls.
By overrding `fetchOpenPrices` method created in #2150

### Linked issue

#230

### DIFF
--- a/src/services/openPricesApi.js
+++ b/src/services/openPricesApi.js
@@ -74,7 +74,7 @@ function extraPriceCreateOrUpdateFiltering(data) {
  * 2. to ensure cookies are not sent to the API. Why?
  * - Open Prices backend API allows both cookie & token authentication
  * - but Open Prices frontend (this app) only uses token authentication
- * - sending both can lead to issues (the cookie likely coming from another OFF domain or env)
+ * - sending both can lead to issues (e.g. the cookie coming from Django admin)
  */
 function fetchOpenPrices(endpointWithParams, options) {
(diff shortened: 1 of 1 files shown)


src/services/openPricesApi.js
```diff
@@ -68,10 +68,20 @@ function extraPriceCreateOrUpdateFiltering(data) {
...
 }
 
-
+/**
+ * Wrapper around fetch
+ * 1. to avoid repeating VITE_OPEN_PRICES_API_URL
+ * 2. to ensure cookies are not sent to the API. Why?
+ * - Open Prices backend API allows both cookie & token authentication
+ * - but Open Prices frontend (this app) only uses token authentication
+ * - sending both can lead to issues (the cookie likely coming from another OFF domain or env)
```
TTalex: Offending cookie comes from Django Admin.

TTalex (approved):
Should work :)

src/services/openPricesApi.js
raphodn: clarified in https://github.com/openfoodfacts/open-prices-frontend/pull/2148/commits/5d71ee8e1c795226fcb6c514a5966fa1fc1e3e28 :+1:

PR #2154, opened 2026-04-17
refactor(LocationSelectorDialog): remove 'Results' title and show the list of cards directly (like the BarcodeScannerDialog)

raphodn:
### What

src/components/LocationSelectorDialog.vue
- remove some (useless) text
- add margin between the form and the results

### Screenshot

|Before|After|
|-|-|
|image|image|

### DIFF
--- a/src/components/LocationSelectorDialog.vue
+++ b/src/components/LocationSelectorDialog.vue
@@ -42,7 +42,7 @@
           </v-tabs-window-item>
 
           <v-tabs-window-item value="osm">
-            <v-form @submit.prevent="locationOsmSearch">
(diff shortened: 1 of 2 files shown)


src/components/LocationSelectorDialog.vue
```diff
@@ -67,41 +67,33 @@
               </i18n-t>
             </p>
 
-            <v-sheet v-if="results !== null">
```
raphodn, line 70: I went a bit too fast on this one, I fixed a bit the display and will do another PR tomorrow to get back the behavior (avoid displaying messages on load)

PR #2161, opened 2026-04-17
refactor(Price Card): when product is missing, turn default image to red

### LINKED ISSUE #2163: Better icons to help differenciate products
### Story

We currently have 4 different "levels" of products

|Description|Image (currently)|Idea|
|-|-|-|
|Barcode product (in OxF) with image|Image||
|Barcode product (in OxF) without image|Image||
|Barcode product unknown|Image||
|Category product (taxonomy)|Image|find a more fruit-vegetable image ?|

In #2161 we started differentiating known & unknown barcode products (changing the icon color to red).
But I feel there is still improvements...

### Todo

- [x] differentiate known & unknown barcode products: #2161
- [ ] find another image for category products ?

raphodn:
### What

Currently there's little difference between an existing product without an image, and an unknown product (apart from the `ProductMissingChip` and maybe the name set to the barcode).

The idea here is to change the default icon color

### Thoughts

There's probably better to do here, for instance an alternate image ?

### Screenshot

|Before|After|
|-|-|
|image|image|

in https://github.com/openfoodfacts/open-prices-frontend/pull/2161/commits/88f2acbb1ccdb7a2e5f10903040fb6b0c36a79ee the color was improved

### DIFF
--- a/src/components/PriceCard.vue
+++ b/src/components/PriceCard.vue
@@ -5,7 +5,7 @@
         <v-col v-if="!hideProductImage" class="pr-0" style="max-width:20%;">
           <v-img v-if="product && product.image_url" :src="product.image_url" max-height="100px" @click="goToProduct()" />
           <v-img v-else-if="product && product.source || price.category_tag" :src="productImageDefault" width="100px" style="filter: invert(0.9);" />
-          <v-img v-else :src="productImageDefault" width="100px" style="filter: invert(25%) sepia(50%) saturate(2000%) hue-rotate(350deg);" />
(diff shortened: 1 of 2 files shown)


TTalex:
Could use another image, like a question mark ?

If we go the coloring way, it would be best if the color matches the Product Missing tag color.

Can use : https://codepen.io/sosuke/pen/Pjoqqp for filter computation.

This one looks decent in dark mode
```
filter: invert(57%) sepia(22%) saturate(6809%) hue-rotate(314deg) brightness(96%) contrast(68%);
```

raphodn:
You rock, i was way too lazy to understand the filter options ^^

Regarding the image, we probably should have a few
- barcode product with image: show image
- barcode product without image: show the current default in white ?
- barcode product missing: show the current default in red? Or a question mark? Or something else?
- category product: another (more "bulk") image ?

I'll merge as is and open an issue

PR #2162, opened 2026-04-17
refactor(CreateOffProduct): improve barcode input UI (full width)

raphodn:
### What

Simplify the barcode input
- full width
- reduce height

### Why

to give more width to the product list below, and in a futur PR
- load less prices
- start displaying images

### Screenshot

|Before|After|
|-|-|
|image|image|


TTalex (approved):
No strong feelings on that one. 👍 

Just know that it probably looked cramped because you have the "Display barcode" option selected.

PR #2185, opened 2026-05-13
fix: keep en:breads in category list

Freso:
“Breads” is a broad category and there will not always be a better more specific category for an item, so it would be good to still have generic “Breads” in the list of categories to choose from.

Needed-for: https://prices.openfoodfacts.org/proofs/112725


TTalex (approved):
Makes sense to me to have it as a default value in case no specific bread category can be found.

raphodn:
fyi I see only 1 file changed, which means the PR didn't run `generate_categories_json_per_language` to update the JSON. but it's fine we'll wait for this weekend for the CRON to run and update everything at that moment :)

PR #2195, opened 2026-05-16
refactor(Product Detail): if unknown, show only 1 banner to add the product

### LINKED ISSUE #2076: Improve layout of options on the "Product not found" screen
### What

When a product is not found, the interface shows the following elements close together:

- Add to Open Food Facts
- Experiment: create this product using data from Open Prices

Both appear as prominent actions and seem to solve the same problem (adding the missing product). This can be confusing because it is not clear which action users should take or how they differ.

A cleaner & nicer layout could make it easier to understand.

### Steps to reproduce the behavior:
1. Add a product barcode that is not present in Open Food Facts. 
2. View the "Product not found" screen. 
3. Observe the two available actions. 

### Expected behavior

The "product not found" screen should present the available options in a clearer and more visually organized way.

The actions **Add to Open Food Facts** and **Create this product using data from Open Prices (experiment)** should be grouped or styled so it is easier to understand that they are two different ways to create the product.

This would make the screen easier to scan and help users quickly understand the available options.

### Screenshots

Image

### Part of
-

raphodn:
### What

Following #1959 and the creation 
For an unknown product, improve the display of the banners pushing to create it.

Closes #2076

### Screenshot

||Before|After|
|-|-|-|
|product with 0 prices|image|same|
|product with at least 1 price|image|image|

### DIFF
--- a/src/views/ProductDetail.vue
+++ b/src/views/ProductDetail.vue
@@ -8,8 +8,10 @@
 
   <v-row v-if="productOrCategoryNotFound" class="mt-0">
     <v-col cols="12" sm="6">
-      <ProductNotFoundAlert v-if="productNotFound" :productCode="productId" />
-      <CreateOpenFoodFactsProductPromoBanner v-if="productNotFound && priceTotal" class="mt-3" :productCode="productId" />
+      <template v-if="!productNotFound">
+        <CreateOpenFoodFactsProductPromoBanner v-if="priceTotal" class="mt-3" :productCode="productId" />
+        <ProductNotFoundAlert v-else :productCode="productId" />
+      </template>
       <CategoryNotFoundAlert v-else-if="categoryNotFound" :categoryTag="productId" />
     </v-col>
   </v-row>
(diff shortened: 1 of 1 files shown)


src/views/ProductDetail.vue
```diff
@@ -8,8 +8,10 @@
 
   <v-row v-if="productOrCategoryNotFound" class="mt-0">
     <v-col cols="12" sm="6">
-      <ProductNotFoundAlert v-if="productNotFound" :productCode="productId" />
-      <CreateOpenFoodFactsProductPromoBanner v-if="productNotFound && priceTotal" class="mt-3" :productCode="productId" />
+      <template v-if="!productNotFound">
```
raphodn, line 11: oops, typo..

raphodn: fixed here: https://github.com/openfoodfacts/open-prices-frontend/commit/a76bf10eb1282789f4cf3614d15233f08d90f5d8

PR #2207, opened 2026-05-24
fix(ui): hide unknown brand chip when product is in readonly mode

### LINKED ISSUE #1409: Price card shows "? brand" badge when brand is unknown
## Description
On the prices/search page, when a product has no brand information, 
the card displays a badge with a literal "? brand" text in orange. 
This looks broken and confusing to users.

## Steps to Reproduce
1. Go to the prices page on the explorer
2. Look for any product card that has no brand data (e.g. "Iced Latté")
3. The brand badge shows "? brand" instead of being hidden or showing "Unknown"

## Expected Behavior
Either:
- Hide the brand badge entirely if no brand is available, OR
- Show a neutral "Unknown brand" label without the question mark styling

## Screenshots

Image

## Environment
- Browser: brave
- Page: Prices / search results page

ishikajain-234:
## Problem
When a product has no brand, a "? brand" chip is displayed 
with a warning color and help icon. This looks broken and 
confusing to users browsing the app.

## Fix
Hide the unknown brand chip when the component is in readonly 
mode (i.e. when just browsing, not editing).
## Related Issue
Fixes openfoodfacts/openfoodfacts-explorer#1409

### DIFF
--- a/src/components/ProductBrands.vue
+++ b/src/components/ProductBrands.vue
@@ -5,7 +5,7 @@
         {{ brand }}
       </v-chip>
     </template>
-    <v-chip v-else label size="small" density="comfortable" @click="showProductBrandsDialog">
+    <v-chip v-else-if="!readonly" label size="small" density="comfortable" @click="showProductBrandsDialog">
       <i>{{ $t('ProductCard.BrandTotal', { count: productBrandsList.length }) }}</i>
     </v-chip>
     <ProductBrandsDialog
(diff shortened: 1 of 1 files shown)


DevDs1989:
I don't think this is an issue that needs to be fixed.The `? brand` tag signals that an important piece of data is missing from OFF and that shows urgency and prompts the user implicitly to contribute and help enrich the data.

Hiding it in read only mode would take away critical information and in turn take away this contribution opportunity.

@raphodn what are your thoughts on this?

raphodn:
> I don't think this is an issue that needs to be fixed

Indeed this hasn't been raised in an issue nor discussed. 
For now, and to simplify the maximum the UI, i would keep the display of the PriceCard & ProductCard as homogenous as possible in the different pages & context :pray: 

Alternatively, "readonly" could still display a menu and allow opening the product in another tab to edit it in OFF ?

PR #2254, opened 2026-07-05
feat(ProofUpload): Upload as drafts first

TTalex:
### What
- Proposed implementation of the proof draft system
- As soon as proofs are selected, they are uploaded as drafts
- Revamped the proof displaying list, to show drafts currently being uploaded
- Replaced old upload button with drafts finalize

### Screenshot
Screenshot 2026-07-05 at 16-24-44 Add proofs (price tags) Open Prices

### Notes
Fundamentally, using drafts do not currently give any benefits to users, so merging this PR might not be the best idea until said benefits are figured out.

### DIFF
--- a/src/components/ProofUploadCard.vue
+++ b/src/components/ProofUploadCard.vue
@@ -26,20 +26,21 @@
         </v-row>
         <ProofTypeInputRow :class="showTopAlertOrBanner ? 'mt-0' : ''" :proofTypeForm="proofForm" :typePriceTagOnly="typePriceTagOnly" :typeReceiptOnly="typeReceiptOnly" />
         <LocationInputRow class="mt-0" :locationForm="proofForm" @location="locationObject = $event" />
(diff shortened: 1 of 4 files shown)


src/components/ProofUploadCard.vue
raphodn, line 1: an idea could be to restrict the draft workflow to receipt proofs to start with ?

raphodn: possibly duplicating the ProofUploadCard and having a ProofDraftUploadCard (or ProofReceiptUploadCard) and thus being more flexible in UI tweaks

TTalex: Agreed, closing this PR for now.

I'll update https://github.com/openfoodfacts/open-prices-frontend/pull/2255 to have something specific to the receipt flow that limits edits to the proof upload component.

PR #2314, opened 2026-08-01
fix(Price Card): better manage when price currency is missing

raphodn:
### What

The backend allows proofs to be created without date & currency. Sometimes (rarely) prices also have these fields missing.
This PR avoids the frontend to crash if it is the case.
- better manage price "locale display" when currency missing
- show a "currency missing chip" if it is the case

### Screenshot

|Before|After|
|-|-|
|image|image|


teolemon:
what is the 🍎 icon about ?

raphodn:
> what is the 🍎 icon about ?

it's in the settings you can optionally display from which flavor the product comes from (oxf)

PR #2330, opened 2026-08-13
refactor(SEO): update more metatags on route change

### LINKED ISSUE #914: SEO: Dynamic metadata depending on the page
### Story

Currently, when we share an OP url, it is always the same metadata tagline that is displayed.

But when we share the URL of a product, a location, or a "top page", it should be more explicit and make users want to click :)

### Screenshot

Image

### Tools

- https://metacheck.appstate.co
- https://metatags.io/

raphodn:
### What

In #2189 we already had dynamic title & description (linked to issue #914)
Here we also make dynamic the og:title & twitter:title (same for description)

### How

I first started with https://github.com/unjs/unhead
- but it works per component/page, and we are doing it at the router level
- i struggled/didn't find how to import useHead, it wasn't created yet in the router it seems
- reverted to our current strategy, with a new `setMeta` method

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -120,6 +120,20 @@ function replaceStringWithList(value) {
 }
 
 /**
+ * Create or update a meta tag by attribute name/value.
+ */
+function setMeta(attrName, attrValue, content) {
+  const selector = `meta[${attrName}="${attrValue}"]`
+  let el = document.querySelector(selector)
+  if (!el) {
+    el = document.createElement('meta')
+    el.setAttribute(attrName, attrValue)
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -119,6 +119,20 @@ function replaceStringWithList(value) {
   return value
 }
 
+/**
+ * Create or update a meta tag by attribute name/value.
+ */
+function setMeta(attrName, attrValue, content) {
```
raphodn, line 125: i'll merge and we'll see if it improves anything 🤷

PR #2332, opened 2026-08-16
feat(Prices): new nearby filter (lat, lon & radius_km)

### LINKED ISSUE #277: Prices API: new filter for locations within n meters, after providing latitude/longitude

TaciteOFF:
> [!WARNING]
> **Depends on openfoodfacts/open-prices#1419** (which closes openfoodfacts/open-prices#277). Please do not merge this before that PR is merged **and deployed**.
>
> Until the API supports `lat`, `lon` and `radius_km`, it ignores them and answers 200 with every price, while the filter chip still shows as active. Nothing errors out, nothing looks empty, so the bug is invisible. The e2e tests cannot catch it either, since they stub the API. I keep this PR as a draft until then.

### What it does

A new filter chip on the price list and on product pages restricts prices to a radius around either the current position or a place picked from the location selector.

The center and the radius live in the URL, so a filtered view stays shareable and bookmarkable, and the back button behaves.

The radius offers 1, 2, 5, 10, 20, 50 and 100 km. The API sets no maximum, so this list is only a UI choice. The largest option is also the bound the URL parser accepts, so a link asking for more is ignored instead of being half applied.

### Location selector: physicalOnly

Online locations have no coordinates, so they cannot be a search center. The selector gets a `physicalOnly` prop that hides the Online tab and keeps only the favourites and recents that do have coordinates.

That filtering exposed a data-loss bug worth mentioning. `Clear` called `clearFavoriteLocations()` in every case. So a user seeing 2 physical favourites and clicking Clear would have deleted all 5 of their stored favourites, online ones included. In `physicalOnly` mode it now removes only what the dialog shows.

### Shared mixin

`initPrices` and `getPrices` were duplicated between `PriceList.vue` and `ProductDetail.vue`. They move to `src/mixins/priceList.js`, with two hooks (`beforeInitPrices`, `onPricesLoaded`) for the parts that really differ. Both views get an error state with a retry, and two fixes come with it:

- the page counter only moves forward once a page has really landed, so a failed request no longer skips a page;
- infinite scroll no longer replays a request that is showing an error, which used to hammer a failing endpoint on every scroll.

Seven other views still carry the same duplicated pair. They could use the mixin later, that felt out of scope here.

### Tests

`yarn lint` reports **0 errors**, and the `lint-staged` hook passes `--max-warnings=0` on every file of this PR.

Three e2e tests cover the feature: the full geolocation journey, checking that `lat`, `lon` and `radius_km` reach `/prices` and travel with `date__gte` **in a single request**; a radius above the maximum being ignored; and the retry path.

The e2e job is commented out in CI ("tests are not up to date"), so I ran them locally against a database loaded with the production dumps, 290 792 prices. On top of the stubbed suite, I wrote a temporary spec with no stubbing at all, to drive the real app against the real API. It confirmed the totals I had computed separately from the raw JSONL: 1 214 prices for Paris / 1 km, 79 545 for 20 km, 290 792 with no filter, and 21 540 for 5 km, going down further once the date filter is added. I removed that spec afterwards, since a test depending on real database content is exactly what makes the other ones fail today.

One fix found while running the suite: `FilterMenu` is a `v-menu` with `scroll-strategy="close"`, so the scroll Cypress does before a click closed the menu and detached the item. The click now passes `scrollBehavior: false`.

### Note

`.gitignore` gains `.DS_Store`. Unrelated to the feature, happy to split it out if you prefer.


TaciteOFF:
Warning: When using "use my position", toFixed(5) gives 5 digits precision (~1m).
matomo sends pages URL to analytics.openfoodfacts.org 
So it could be possible for admins to link IP to GPS position.  

It could be wise to : change to 4 digits + setExcludedQueryParams in Matomo

src/mixins/priceList.js
```diff
@@ -0,0 +1,60 @@
...
+      this.priceTotal = null
+      this.pricePage = 0
+      this.priceListError = ''
+      this.loading = false
+      return this.getPrices(requestId)
+    },
+    getPrices(requestId = this.priceRequestId) {
+      // the scroll handler calls this too: don't stack requests, and don't
+      // retry on its own once an error is displayed (see retryPrices)
+      if (requestId !== this.priceRequestId || this.loading || this.priceListError) return
```
raphodn: could you move this to another PR ? i'm not sure i understand completely yet, but in any case its not linked to geo filtering right ?

src/mixins/priceList.js
raphodn, line 1: it'll probably be hard to undo, but i would have created the mixin in another PR.
when searching for "priceList: []," i get over 10 results, there's quite a few pages where it could be added, it could have been a seperate refactoring PR 👀

src/components/NearbyPriceFilter.vue
```diff
@@ -0,0 +1,230 @@
...
+  <v-dialog v-model="dialog" :width="dialogWidth">
+    <v-card :title="$t('NearbyPriceFilter.Title')">
+      <template #append>
+        <v-icon icon="mdi-close" :aria-label="$t('Common.Close')" @click="closeDialog" />
+      </template>
+
+      <v-divider />
+
+      <v-card-text>
+        <v-alert
```
raphodn: once the location is selected, rather display it under the 2-btn row ?

src/components/NearbyPriceFilter.vue
```diff
@@ -0,0 +1,230 @@
...
+            </v-btn>
+          </v-col>
+          <v-col cols="12" sm="6">
+            <v-btn
+              data-name="nearby-price-choose-area-button"
+              block
+              prepend-icon="mdi-map-search-outline"
+              @click="locationSelectorDialog = true"
+            >
+              {{ $t('NearbyPriceFilter.ChooseArea') }}
```
raphodn:
```suggestion
              {{ $t('NearbyPriceFilter.ChooseLocation') }}
```

raphodn: i'm not sure it's useful to introduce a new word, when "location" works fine here :)

src/components/NearbyPriceFilter.vue
raphodn, line 1: not necessarily something to solve in this PR, but for future discussion: the position of this filter.
at first i thought it should sit under "Filter".
but maybe it could be an "app wide" filter in the long run ? something that is toggled elsewhere than in the list section 🤔

src/components/NearbyPriceFilter.vue
```diff
@@ -0,0 +1,230 @@
...
+          v-if="hasCurrentFilter"
+          data-name="nearby-price-filter-clear-button"
+          color="error"
+          variant="text"
+          @click="clearFilter"
+        >
+          {{ $t('Common.Clear') }}
+        </v-btn>
+        <v-spacer />
+        <v-btn variant="text" @click="closeDialog">
```
raphodn: in other dialogs, we've made the choice that close is already possible in the top right and hitting "Esc", so no need to add an extra button in the footer 🙏

src/utils/geo.js
```diff
@@ -215,6 +222,55 @@ function getLocationOSMLatLng(locationObject) {
...
+function hasLocationCoordinates(locationObject) {
+  const [lat, lon] = getLocationOSMLatLng(locationObject)
+  return toFiniteNumber(lat) !== null && toFiniteNumber(lon) !== null
+}
+
+function getNearbyFilter(query) {
+  const lat = toFiniteNumber(query.lat)
+  const lon = toFiniteNumber(query.lon)
+  const radius_km = toFiniteNumber(query.radius_km)
+  if (
```
raphodn, line 247: alternatively return { lat, lon, radius_km } only when all of the conditions are met ? 😅 
and/or make intermediate conditions per item
(can be refactored in another PR)

src/components/LocationSelectorDialog.vue
```diff
@@ -175,6 +175,12 @@ export default {
     LocationCard: defineAsyncComponent(() => import('../components/LocationCard.vue')),
     LeafletMap: defineAsyncComponent(() => import('../components/LeafletMap.vue')),
   },
+  props: {
+    physicalOnly: {
```
raphodn, line 179: i see what you're doing here 😁 
i think i would have taken a different/simpler approach
- prop hideLocationOnlineTab (similar to BarcodeScannerDialog)
- prop locationOnlineReadonly (to avoid the if/else). one (small) caveat would be that some physical locations might not have lat & lon, but that's a border case, or even just a non existent case ?

raphodn: i don't know honestly we can leave like this, and when we'll need an additional feature/filter on the LocationSelector we can refactor things differently

src/mixins/priceList.js
TaciteOFF: The request ID guard is useful when the URL filter changes during an active request, but, yes, most of the retry and error handling is not specific to geo filtering. I will move this.

src/mixins/priceList.js
TaciteOFF: Agreed. This PR will stay focused on nearby filtering!

src/components/NearbyPriceFilter.vue
TaciteOFF: Good idea,, might be easier to read.

src/components/NearbyPriceFilter.vue
TaciteOFF: Make sense! I'll rename this

src/components/NearbyPriceFilter.vue
TaciteOFF: I kept it close to the price list because its state currently belongs to the page URL. An app-wide location filter could be useful later, especially if more views start using it. We can decide that in another PR?

src/components/NearbyPriceFilter.vue
TaciteOFF: Noted!

src/utils/geo.js
TaciteOFF: I’ll return { lat, lon, radius_km } only when all three values are valid and within their bounds, then return null otherwise

src/components/LocationSelectorDialog.vue
TaciteOFF: I’ll keep physicalOnly for this PR. It also prevents selecting favorites or recent locations without coordinates.

raphodn (approved):

PR #2360, opened 2026-08-28
fix(Lists): don't crash when the API returns an error payload

### LINKED ISSUE #2344: TypeError: t.items is not iterable (cannot read property undefined)

AchrafReyani:
Follow-up to #2357 (#2344, Sentry `OPEN-PRICES-FRONTEND-18`), as discussed in https://github.com/openfoodfacts/open-prices-frontend/pull/2357#discussion_r3879989506.

## Problem

Every other paginated list view does `this.xList.push(...data.items)` unconditionally. When the API answers with an error payload that has no `items` — `404 {"detail":"Invalid page."}` for a page past the last one (easy to hit: the scroll handlers have no in-flight guard, so scrolling while the last page is loading requests the next one), or `400 {"detail":"Maximum page reached…"}` for page ≥ 1000 — the spread throws and `loading` stays `true`.

### Reproduction (before)

Local proxy in front of the real API that answers any `?page=2` request with the API's real `404 {"detail":"Invalid page."}` body, `npm run dev`, scroll to the bottom of the page:

```
GET /api/v1/prices?page=2&size=25&order_by=-created -> 404 {"detail":"Invalid page."}
GET /api/v1/products?page=2&size=25&order_by=-price_count -> 404 {"detail":"Invalid page."}
GET /api/v1/products?page=2&size=25&categories_tags__contains=en%3Abreakfast-cereals&order_by=-price_count -> 404 {"detail":"Invalid page."}
```

```
TypeError: data.items is not iterable (cannot read property undefined)   at src/views/PriceList.vue:78
TypeError: data.items is not iterable (cannot read property undefined)   at src/views/ProductList.vue:74
TypeError: data.items is not iterable (cannot read property undefined)   at src/views/CategoryDetail.vue:73
```

## Fix

Same guard `BadgeDetail`, `BrandDetail`, `LocationDetail`, `ProofDetail` and `UserBadgeList` already use, applied to the 26 remaining call sites: `this.loading = false`, then `if (!data.items) return` before the spread. `ProductDetail` and `UserDetail` had the guard commented out — enabled it. No other logic changed (`loading = false` just moves to the top of the `.then`, where it already is in the guarded views).

## Verification

- Same scenario after the change on `/prices`, `/products`, `/categories/en:breakfast-cereals`, `/proofs`, `/locations`: the proxy still logs the page-2 `404`s, the console has no exception, the 25 loaded items stay rendered and the spinner disappears.
- `npx eslint <changed files>`: clean. `npm run lint`: 0 errors (42 pre-existing warnings).

### DIFF
--- a/src/components/UserRecentProofsDialog.vue
+++ b/src/components/UserRecentProofsDialog.vue
@@ -121,9 +121,10 @@ export default {
       this.userProofPage += 1
       return openPricesApi.getProofs(this.getProofParams)
         .then((data) => {
+          this.loading = false
+          if (!data.items) return
           this.userProofList.push(...data.items)
           this.userProofTotal = data.total
-          this.loading = false
         })
         .catch((error) => {
           console.error(error)
--- a/src/views/BadgeList.vue
+++ b/src/views/BadgeList.vue
@@ -60,9 +60,10 @@ export default {
       this.badgePage += 1
       openPricesApi.getBadges(this.getBadgesParams)
         .then((data) => {
+          this.loading = false
+          if (!data.items) return
           this.badgeList.push(...data.items)
           this.badgeTotal = data.total
-          this.loading = false
         })
     },
   },
--- a/src/views/CategoryDetail.vue
+++ b/src/views/CategoryDetail.vue
@@ -103,9 +103,10 @@ export default {
       this.categoryProductPage += 1
       return openPricesApi.getProducts(this.getProductsParams)
         .then((data) => {
+          this.loading = false
+          if (!data.items) return
           this.categoryProductList.push(...data.items)
           this.categoryProductTotal = data.total
-          this.loading = false
         })
     },
     updateFilterList(newFilterList) {
--- a/src/views/ChallengeList.vue
+++ b/src/views/ChallengeList.vue
@@ -121,9 +121,10 @@ export default {
       this.challengePage += 1
       return openPricesApi.getChallenges(this.getChallengesParams)
         .then((data) => {
+          this.loading = false
+          if (!data.items) return
           this.challengeList.push(...data.items)
           this.challengeTotal = data.total
-          this.loading = false
         })
     },
     handleScroll(event) {  // eslint-disable-line no-unused-vars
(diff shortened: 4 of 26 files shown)


raphodn (approved):

src/components/UserRecentProofsDialog.vue
```diff
@@ -121,9 +121,10 @@ export default {
       this.userProofPage += 1
       return openPricesApi.getProofs(this.getProofParams)
         .then((data) => {
+          this.loading = false
```
raphodn, line 124: what do you (or Claude ^^) think regarding refactoring this in a [mixin](https://v2.vuejs.org/v2/guide/mixins.html) ? useful or useless ?

AchrafReyani: Honest take: probably not worth it as a mixin. The guard is one line and every caller does something different right after it (different list/total/loading names), so a mixin would only wrap `if (!data.items) return` and hide the early return inside a helper. There are also no mixins in the codebase yet, and Vue 3 tends to push composables instead.

Where sharing would really pay off is one level down: make `fetchOpenPrices` (openPricesApi.js) reject when `!response.ok`, so a 404 `{"detail": "Invalid page."}` goes to the `.catch()` every component already has, and the per-component guards become belt and braces. That is a behaviour change for callers that currently read error bodies from a resolved promise (form submits showing `detail`, etc.), so it would need a look at each caller — happy to do that in a separate PR if you'd like.

raphodn: I'm not 100% I see the impact but yes go ahead and we'll see what the PR looks like :)

AchrafReyani: Done in #2368: `fetchOpenPrices` now rejects with an `OpenPricesApiError` (status + payload) on non-2xx responses, and the callers that read error payloads from the resolved promise are updated so nothing visible changes. Two commits: a mechanical `.finally()` one for the loading flags first, then the actual change — the PR description has the details and a before/after table.

PR #2364, opened 2026-08-29
fix(Price List): fix price type filter management in 2 pages

raphodn:
### What

In `LocationDetail` & `UserDetail`, the filter management wasn't fully implemented


src/views/LocationDetail.vue
```diff
@@ -144,6 +149,11 @@ export default {
       this.$router.push({ query: { ...this.$route.query, [constants.FILTER_PARAM]: this.currentFilterList } })
       // this.initPrices() will be called in watch $route
     },
+    updateCurrentType(typeKey) {
```
raphodn, line 152: @TaciteOFF maybe this could be come a mixin ?

PR #2368, opened 2026-08-30
refactor(API): reject with an OpenPricesApiError when the API answers with a non-2xx status

AchrafReyani:
Follow-up to #2360, as discussed in https://github.com/openfoodfacts/open-prices-frontend/pull/2360#discussion_r3883868058.

### Problem

`fetchOpenPrices` resolves with `response.json()` whatever the HTTP status, so an API error like `{"detail": "Maximum page reached. …"}` reaches the callers' `.then()` as if it were data. That is why the list views needed the `if (!data.items) return` guards (#2360), why form submits sniff `data.id`, why the sign-in view sniffs `data.access_token`, and why every other call site silently swallows API errors (nothing in the console, nothing in Sentry).

### Fix (2 commits, best reviewed separately)

1. **`refactor(API): reset loading flags in .finally() instead of .then()`** — mechanical, no behaviour change on success. Every `openPricesApi` chain that reset a `loading` flag inside `.then()` (and had no `.catch()`) now does it in `.finally()`. Needed so that the loading flags don't stay stuck once errors reject (44 files, one hunk each, applied by a script and reviewed by hand).

2. **`fix(API): reject with an OpenPricesApiError when the API answers with a non-2xx status`** — `fetchOpenPrices` now rejects with an `OpenPricesApiError` (`status`, `data` = parsed payload, `message` = the payload's `detail`) on any non-2xx response, so API errors flow to `.catch()` like network errors already did. The callers that read error payloads from the resolved promise are updated so that the user-visible behaviour stays the same:
   - `SignIn`: 401 → "wrong credentials" alert (was: no `access_token` in the body)
   - `getProductByCode` callers (`BarcodeScannerDialog`, `ProductInputRow`, `ReceiptTableCard`, `ProductDetail`, `CreateOffProduct`): 404 → placeholder product, as before
   - `LocationOSMDetail`: 404 → location built from the OSM type & id, as before
   - price/proof create & upload forms: the alert now comes from `.catch()` and shows the API's `detail` (e.g. `Error: Authentication credentials were not provided.` instead of `Error: {"detail":"…"}`); the `else` branches that displayed it are gone, and their `loading` flags are reset in `.finally()`

The `if (!data.items) return` guards from #2360 are kept (belt and braces).

**Impact to be aware of:** call sites without a `.catch()` (mostly the list views) now surface API errors as an unhandled rejection — `OpenPricesApiError: Maximum page reached. …` in the console, and in Sentry in prod — instead of ignoring them. The UI stays stable (loading flags reset, items already displayed are kept). If you'd rather keep some of them quiet, tell me which and I'll add a `.catch()`.

### Verification

Local setup: `.env.local` pointing `VITE_OPEN_PRICES_API_URL` to a small Node proxy on `127.0.0.1:8000` that forwards to the live API but answers any `?page>=2` request with the real API's `400 {"detail":"Maximum page reached. …"}`; `vite --port 5180`; driven in Chrome with `window.alert` replaced by a logger and an `unhandledrejection` listener.

| scenario | before (`main` @ 7f46ea17) | after (this branch) |
|---|---|---|
| `/prices`, scroll to the bottom → page 2 answered with 400 | request made, **nothing** logged, `loading=false`, 25 items | `OpenPricesApiError status=400 "Maximum page reached. …"` as an unhandled rejection, `loading=false` (via `.finally()`), no spinner, 25 items kept |
| `/sign-in`, wrong credentials (real `POST /auth` → 401) | alert `Error: wrong credentials` | alert `Error: wrong credentials` |
| `/prices/add/single`, barcode `12345678901234567890` (real `GET /products/code/…` → 404) | `productForm.product = {"code":"…","price_count":0}` | same |
| `/prices/add/single`, submit with an invalid token (real `POST /prices` → 403) | alert `Error: {"detail":"Authentication credentials were not provided."}`, `loading=false` | alert `Error: Authentication credentials were not provided.`, `loading=false` |
| `/products/12345678901234567890` (404) | — | `product = {"code":"…","price_count":0}`, "Unknown product" card rendered, no rejection |
| `/locations/osm/node/1` (404) | — | `location = {"osm_type":"node","osm_id":"1"}`, no rejection |

`npx eslint --max-warnings=0` on the 52 changed files: clean. `npm run lint`: 0 errors (warnings are pre-existing).

Related: the Sentry crash in #2343 (`SyntaxError: Unexpected end of JSON input` in the openPricesApi bundle) is most likely this same problem surfacing on an error response with an empty body — with this PR, non-2xx bodies are parsed with a `.catch(() => null)` guard, so that path no longer throws a SyntaxError.

### DIFF
--- a/src/components/BarcodeScannerDialog.vue
+++ b/src/components/BarcodeScannerDialog.vue
@@ -253,8 +253,11 @@ export default {
       }
       openPricesApi
         .getProductByCode(code)
-        .then((data) => {
-          const product = data.id ? data : {'code': code, 'price_count': 0}
+        .catch((error) => {
+          if (error.status === 404) return {'code': code, 'price_count': 0}  // product not in Open Prices (yet)
+          throw error
+        })
+        .then((product) => {
           if (search) {
             this.productSearchResultList.push(product)
           } else {
--- a/src/components/ProductInputRow.vue
+++ b/src/components/ProductInputRow.vue
@@ -221,8 +221,12 @@ export default {
       this.productForm.product = null
       openPricesApi
         .getProductByCode(code)
-        .then((data) => {
-          this.productForm.product = data.id ? data : {'code': code, 'price_count': 0}
+        .catch((error) => {
+          if (error.status === 404) return {'code': code, 'price_count': 0}  // product not in Open Prices (yet)
+          throw error
+        })
+        .then((product) => {
+          this.productForm.product = product
         })
         .catch((error) => {
           alert(this.$t('Common.ServerError'))
--- a/src/components/ProofUploadCard.vue
+++ b/src/components/ProofUploadCard.vue
@@ -363,16 +363,11 @@ export default {
               .createProof(proofImageCompressed, this.proofForm, this.$route.path)
               .then((data) => {
                 this.loading = false
-                if (data.id) {
-                  resolve(data)
-                } else {
-                  alert(`Error: ${JSON.stringify(data)}`)
-                  console.log(JSON.stringify(data))
-                }
+                resolve(data)
               })
               .catch((error) => {
-                alert(`Error: ${JSON.stringify(error)}`)
-                console.log(JSON.stringify(error))
+                alert(`Error: ${error.message}`)
+                console.log(error)
                 this.loading = false
               })
           })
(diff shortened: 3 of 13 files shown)


src/components/LocationSelectorDialog.vue
raphodn, line 1: hi! Sorry in advance, but could you do 2 (stacked) PRs ? So that we have the 2 changes in the history (as we squash PRs...). thanks 🙏

AchrafReyani: Done 👍 Opened #2369 with just the mechanical .finally() commit; this PR is stacked on it (same 2 commits, so its diff will shrink to the semantic change once #2369 is squash-merged and I rebase this branch).

raphodn (approved):

